PR #32, opened 2020-03-10
Feature/4 engine-ification

taylorcenters:


src/engine/mod.rs
```diff
@@ -0,0 +1,659 @@
...
+            Error::new(ErrorKind::GraphQLError("INTERNAL SERVER ERROR".to_string()), None)
+        }
+    }
+}
+
+impl WarpgrapherRequestContext for () {
+    fn new() {}
+}
+
+/// A WarpGrapher GraphQL server.
```
shellforge: Need to update the docs (Server -> Engine).

src/engine/mod.rs
```diff
@@ -0,0 +1,659 @@
...
+            }
+        }
+
+        // validation passed
+        Ok(())
+    }
+
+    pub fn execute(
+        &self,
+        req: Json<GraphQLRequest>, //TODO make generic
```
shellforge, line 423: Good idea, we should create a ticket to track this.

tests/server_test.rs
```diff
@@ -2,9 +2,10 @@ mod setup;
```
shellforge: I think we can delete this file.

src/engine/mod.rs
taylorcenters: Docs updated to reflect Server->Engine change. Also undid change we made to disable doc code while dev.

tests/server_test.rs
taylorcenters: Deleted

src/engine/mod.rs
taylorcenters: Issue #33 created for this.

shellforge (dismissed):

shellforge (dismissed):

src/error.rs
```diff
@@ -7,6 +7,10 @@ use std::sync::mpsc::RecvError;
 /// Categories of Warpgrapher errors.
 #[derive(Debug)]
 pub enum ErrorKind {
+    /// Returned by the `From` implementation for a `actix_http::error::Error` object.
+    /// Conversion from `actix_http::error::Error`.
+    ActixHttpError(String),
```
maldrake: Shouldn't this go away if we're stripping out the dependency on a specific HTTP framework?  If it does need to stay, I'd recommend cleaning up the error definition comment.  I can't tell from this description why it would be thrown or what it would mean.

src/error.rs
```diff
@@ -139,6 +143,13 @@ pub enum ErrorKind {
     /// please report it to the warpgrapher team.
     FieldMissingResolverError(String, String),
 
+    /// Returned by the `From` implementation for a `String` object.
+    /// Conversion from `String`.
+    GenericError(String),
```
maldrake: This will be a rough experience for developers using the framework.  I would argue against the existence of a GenericError.  Errors are never generic. There's always something specific that triggers them.

src/error.rs
```diff
@@ -139,6 +143,13 @@ pub enum ErrorKind {
...
+
+    /// Returned by the engine for convenience when converting a json object containing an error message to a string. This error wraps the resulting string.
+    GraphQLError(String),
```
maldrake: I don't quite follow what this really means in concrete terms.  Maybe an example here, in the documentation?

src/engine/mod.rs
```diff
@@ -0,0 +1,657 @@
...
+/// ```
+#[derive(Clone)]
+pub struct Engine<GlobalCtx = (), ReqCtx = ()>
+where
+    GlobalCtx: 'static + Clone + Sync + Send + Debug,
+    ReqCtx: 'static + Clone + Sync + Send + Debug + WarpgrapherRequestContext,
+{
+    pub config: WarpgrapherConfig,
+    pub database: String,
+    pub pool: Option<Pool<CypherConnectionManager>>,
```
maldrake: Does it ever make sense to construct an engine without a database pool?  If not, this probably shouldn't be an option.  If so, then fine as is.

src/engine/mod.rs
```diff
@@ -0,0 +1,657 @@
...
+    pub global_ctx: Option<GlobalCtx>,
+    pub resolvers: WarpgrapherResolvers<GlobalCtx, ReqCtx>,
+    pub validators: WarpgrapherValidators,
+    pub extensions: WarpgrapherExtensions<GlobalCtx, ReqCtx>,
+    pub version: Option<String>,
+    root_node: Option<RootRef<GlobalCtx, ReqCtx>>,
```
maldrake: I don't think we'd ever construct an engine without a root note. This shouldn't be optional.

src/engine/mod.rs
```diff
@@ -0,0 +1,657 @@
...
+    /// let db = Neo4jEndpoint::from_env("DB_URL").unwrap();
+    ///
+    /// let mut engine = Engine::<()>::new(config, db)
+    ///     .build().unwrap();
+    /// ```
+    pub fn new(config: WarpgrapherConfig, database: String) -> Engine<GlobalCtx, ReqCtx> {
+        Engine {
+            config,
+            database,
+            pool: None,
```
maldrake: Per the above, if it doesn't make sense to start an engine without a database pool, this shouldn't ever be None.

src/engine/mod.rs
```diff
@@ -0,0 +1,657 @@
...
+            global_ctx: None,
+            resolvers: HashMap::new(),
+            validators: HashMap::new(),
+            extensions: vec![],
+            version: None,
+            root_node: None,
```
maldrake: Same as per above... this should probably never be None.

src/error.rs
```diff
@@ -156,6 +167,20 @@ pub enum ErrorKind {
     /// defined for a field, but the corresponding resolver is not provided.
     ResolverNotFound(String, String),
 
+    /// This error is returned if the root node could not be determined from the schema.
+    /// Note: This error should never be thrown. This is a critical error. If you see it,
+    /// please report it to the warpgrapher team.
+    MissingRootNode,
```
maldrake: Per the comments on the engine.rs file, I'd recommend looking at ways to get rid of this as a possibility.

src/error.rs
```diff
@@ -156,6 +167,20 @@ pub enum ErrorKind {
...
+
+    /// This error is returned if the database pool was not created.
+    /// Note: This error should never be thrown. This is a critical error. If you see it,
+    /// please report it to the warpgrapher team.
+    MissingDatabasePool,
```
maldrake: Per the comments on the engine.rs file, I'd recommend looking at ways to get rid of this as a possibility.

src/error.rs
```diff
@@ -156,6 +167,20 @@ pub enum ErrorKind {
...
+
+    /// Returned by the `From` implementation for a `serde_json::error::Error` object.
+    /// Conversion from `serde_json::error::Error`.
+    SerdeJsonError(String),
```
maldrake: This tells me about the mechanics of the source of the error, but not really what causes it.  Why would someone get this from Warpgrapher?

src/error.rs
```diff
@@ -170,6 +195,10 @@ pub enum ErrorKind {
     /// Returned when a `WarpgrapherServer` fails to start.
     ServerStartupFailed(RecvError),
 
+    /// Returned by the `From` implementation for a `r2d2::Error` object.
+    /// Conversion from `r2d2::Error`.
+    R2D2Error(String),
```
maldrake: This documentation comment tells me mechanically where the error came from, but not the circumstances under which I might encounter it.  Why would Warpgrapher return such an error?

src/error.rs
```diff
@@ -213,6 +242,30 @@ impl Error {
...
+        Error::new(ErrorKind::R2D2Error(error.to_string()), None)
+    }
+}
+
+impl From<String> for Error {
+    fn from(error: String) -> Error {
+        Error::new(ErrorKind::GenericError(error), None)
+    }
+}
+
```
maldrake: We should be consistent.  For some errors above, we're creating an ErrorKind variant that takes the originating / source error.  In these cases, it looks like we're pulling out strings from errors and just carrying those.  In so doing, are we losing information, like enum variants in the original errors, that would allow for better programmatic handling of conditions?  Even if not, are the strings associated with those errors rich enough in their descriptions to make it easy to debug what's happening?  In some cases, noted above, we probably need better documentation about why these errors would occur in the context of Wrapgrapher.

src/error.rs
```diff
@@ -213,6 +242,30 @@ impl Error {
...
+impl From<r2d2::Error> for Error {
+    fn from(error: r2d2::Error) -> Error {
+        Error::new(ErrorKind::R2D2Error(error.to_string()), None)
+    }
+}
+
+impl From<String> for Error {
+    fn from(error: String) -> Error {
+        Error::new(ErrorKind::GenericError(error), None)
```
maldrake: I think GenericError will probably end up hiding too much information.  It's not necessarily going to be clear why it's being returned when it is.  It seems it will depend almost entirely on the error string in the originating error, which could be very short on useful information.

taylorcenters: Agreed. I got a bit carried away when doing the From impl's

src/engine/mod.rs
taylorcenters: This and the root_node value get constructed in the build() operation since they are not direct pass-through values (they are derived from the database string and the config).

taylorcenters: To address this, will implement an EngineBuilder

src/error.rs
taylorcenters: This has been removed. I believe I added this at one point during development because I abstracted away the enginification away from the resolver tests and am still using Actix there and needed to to propagate out the error.

src/error.rs
taylorcenters: This has been removed

src/error.rs
taylorcenters: This is an error that comes from the pre_request_hook extensions, the error returned from that gets formatted into a GraphQL Json string and used to be returned directly from the Actix server. I created this error to wrap the resulting GraphQL Json string so we can return that from the engine. @shellforge do we have a specific error type coming from extensions that I can pass back directly? Or should I make this a specific error for PreRequestHookExtensionError?

src/engine/mod.rs
```diff
@@ -0,0 +1,697 @@
...
+{
+    pub config: WarpgrapherConfig,
+    pub database: String,
+    pub pool: Option<Pool<CypherConnectionManager>>,
+    pub global_ctx: Option<GlobalCtx>,
+    pub resolvers: WarpgrapherResolvers<GlobalCtx, ReqCtx>,
+    pub validators: WarpgrapherValidators,
+    pub extensions: WarpgrapherExtensions<GlobalCtx, ReqCtx>,
+    pub version: Option<String>,
+    root_node: Option<RootRef<GlobalCtx, ReqCtx>>,
```
shellforge: If I understood yesterday's conversation correctly, part of the motivation to employ `EngineBuilder` was to avoid having these fields as options. 

Instead of passing an `Engine` to `EngineBuilder::new`, perhaps we pass the required args (ex: `WarpgrapherConfig`) and a new `Engine` is only created by `EngineBuilder::build` ?

src/error.rs
taylorcenters: Implemented EngineBuilder to address this

src/error.rs
taylorcenters: Removed this error

src/error.rs
taylorcenters: Wrapped the external errors inside new warpgrapher errors which describe the warpgrapher context. These are no longer used and have been removed.

src/error.rs
taylorcenters: This error has been removed. The external error is now wrapped in a specific Warpgrapher error.

src/engine/mod.rs
taylorcenters: No longer optional

src/error.rs
taylorcenters: Pool and RootNode are no longer Optional parameters for an Engine. These errors have been removed since the condition no longer exists.

src/engine/mod.rs
taylorcenters: No longer Options

src/error.rs
shellforge: PreRequestHookExtensionError sounds appropriate.

shellforge (approved):

src/error.rs
taylorcenters: Replaced GraphQLError with extension specific ErrorKind

PR #34, opened 2020-03-14
Graphson Support

### LINKED ISSUE #1: Fix r2d2 errors
The following logs can be observed when the server is running. Investigate and fix. 

```
[2020-01-29T00:01:03Z ERROR r2d2] Hyper
[2020-01-29T00:01:03Z ERROR r2d2] Hyper
[2020-01-29T00:01:03Z ERROR r2d2] Hyper
[2020-01-29T00:01:03Z ERROR r2d2] Hyper
[2020-01-29T00:01:03Z ERROR r2d2] Hyper
```

### LINKED ISSUE #2: Add CosmosDB and Gremlin Support
- [x] Explore feasibility of using CosmosDB as database

### LINKED ISSUE #12: Review package and API structure and visibility
Right now, almost everything that might be used publicly is exposed through a pub use statement in lib.rs. Look carefully and see whether that flat structure really makes sense, or if it would be easier to learn or use if some of the package structure were exposed.

At the same time, look at whether the current module hierarchy and visibility for modules and structs makes sense, given that it grew more or less organically as we went along. It may make sense to flatten a little at the top and take advantage of pub(crate) and pub(super) in a few places.

### LINKED ISSUE #17: Improve Neo4j driver
- [ ] Add support for Bolt
- [ ] Add support for 4.0

### LINKED ISSUE #28: Single-node rels allows multiple rels

### LINKED ISSUE #29: Reading scalar list props fail

### LINKED ISSUE #33: Update incoming request to be a generic object.
As a legacy of Warpgrapher having the server built-in, the request being operated on is an actix_web::web::Json object. This should be transitioned to a generic Json object (maybe serde_json) to be independent of the service sending the request.

### LINKED ISSUE #37: Revisit error handling
* Adopt "From" pattern as default
* Use map_err when extra context is needed

### LINKED ISSUE #41: Move project-tracker example out of main repo

### LINKED ISSUE #49: Add single database client on request context

maldrake:
So far, this is a work in progress.  Please don't merge yet.  When completed, this branch should provide support for Graphson based databases, in addition to Neo4J.  Along the way, it looks like there will be a redesign of the resolvers and visitors to pull database-specific code out into a set of database modules. This is probably going to be better for longer-term maintainability anyway.


maldrake:
At long last, after much yak shaving, if not bike shedding, this PR is ready for review.  It should close issues #1, #2, #12, #17, #28, #29, #33, #37, #41, and #49.  I didn't intend to tackle so much in one change, and I certain prefer / recommend smaller, bite-sized change packages.  This was just a case of one thing leading to another.  I'll try not do that again.  In summary, for the change log:

- Replaced rusted_cypher with bolt_client, which removed the dependence on r2d2 for Neo4J support, resolving issue #1 as well as #17.  We now support Neo4J up to version 4.1.  (Though possibly no longer less than 4.0. I haven't tested with earlier versions.)
- Added support for ComosDB, resolving #2.
- Reviewed package structure and visibility, splitting things out to make more packages crate-only visible, to simplify the public interface, resolving #12.
- Put in a check on whether a single-rel relationship already exists, resolving #28.
- Corrected scalar property resolver behavior, resolving #29.
- Removed the dependence on actix for a serialization format for incoming requests, moving to a straight serde_json object, resolving #33.
- Revisited error handling to follow guidance from the public API guidelines, fixing #37.  Added as many From conversions for other error objects as practical.
- Removed example projects from the main repo, to avoid cargo feature build conflicts, resolving #41.  We now need to build them out in separate projects.
- Corrected an issue in which we were holding on to clients drawn from the connection pool after a resolver had effectively finished, leading to exhaustion of the pool and timeouts in other resolvers. Now each request only needs one connection at a time.

Please review.  Comments and critique welcome.

README.md
```diff
@@ -24,85 +32,162 @@ See the [Quickstart](https://warpforge.github.io/warpgrapher/warpgrapher/quickst
...
 ```
 
-Run unit tests:
+### Run the Database
+
+For Cosmos DB:
+
+Cosmos DB is an Azure cloud service, so it's already running. Or, if you're using a local Cosmos
+emulator, start the service based on its instructions.
+
```
shellforge, line 91: Let's add a note here about the partition key.

book/src/api/node_delete.md
```diff
@@ -14,13 +14,9 @@ mutation {
                 status: "INACTIVE"
             },
             $DELETE: {
-                force: true
```
shellforge, line 17: @maldrake : was this an error in the docs, or did the force flag get removed?

book/src/engine/context_global.md
```diff
@@ -33,8 +33,8 @@ let engine: Engine<AppGlobalContext, ()> = Engine::new(config, db)
 #### 3. Use GlobalContext in a resolver
 
 ```rust
-fn resolve(context: ResolverContext<AppGlobalContext, ()>) -> ExecutionResult {
-    let global_ctx = context.get_global_context()?;
+fn resolve(facade: ResolverFacade<AppGlobalContext, ()>) -> ExecutionResult {
```
shellforge, line 36: This might be super nitpicky, but I don't love the name `ResolverFacade`. If I was a stranger and saw that name, I think I might be puzzled as to what it meant. Not vetoing it, but I want to think about it.

book/src/warpgrapher/databases.md
```diff
@@ -0,0 +1,54 @@
...
+Add Warpgrapher to your project config:
+
+`cargo.toml`
+
+```toml
+[dependencies]
+warpgrapher = { version = "0.2.0", features = ["cosmos"] }
+```
+
+Then set up environment variables to contact your Cosmos DB:
```
shellforge, line 23: Also a good place to mention the partition key requirement.

src/client/mod.rs
```diff
@@ -1,322 +1,498 @@
...
     ///
     /// # Examples
     ///
     /// ```rust
-    /// use std::env::var_os;
-    /// use warpgrapher::client::Client;
+    /// # use warpgrapher::Client;
     ///
-    /// let mut client = Client::new("http://localhost:5000/graphql");
+    /// let mut client = Client::<(), ()>::new_with_http("http://localhost:5000/graphql");
```
shellforge, line 60: @maldrake : Does rust have any way to specify default type parameters so as to not need to specify two empty objects?

src/client/mod.rs
```diff
@@ -1,322 +1,498 @@
...
+    ///
+    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// let c = Configuration::new(1, Vec::new(), Vec::new());
+    /// let engine = Engine::new(c, DatabasePool::NoDatabase).build()?;
+    ///
+    /// let mut client = Client::<(), ()>::new_with_local(engine);
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn new_with_local(engine: Engine<GlobalCtx, RequestCtx>) -> Client<GlobalCtx, RequestCtx> {
```
shellforge: @maldrake : how would you feel about renaming this to `new_with_engine` ?

book/src/api/node_delete.md
maldrake: I removed the force flag.  In Cypher, the default behavior is to return an error if a query tries to DELETE nodes that still have relationships. The force flag mapped to a DETACH DELETE in Cypher, which forces the deletion of both a node and associated relationships.

Gremlin's behavior is a little different.  Deleting a node automatically deletes the associated relationships, similar to the behavior of DETACH DELETE.  Simulating the behavior of a Cypher DELETE (without DETACH) would require an extra round trip to the database, with additional logic to test for the presence of any relationships before deleting a node.

I decided that it in this case, it was cleaner to support the lowest common denominator of the two databases' behavior, rather than adding a potentially hidden cost to the force flag. If a client wants to check for relationships before deleting a node, it can still do so. It would just need to be done as a separate query.

book/src/engine/context_global.md
maldrake: Please be nit picky. I have no ego about code.

I went with the name `ResolverFacade` because the intent of the code seemed to be to provide a simplified developer experience over raw interaction with the surrounding Warpgrapher code, thus matching the [definition of a facade as a design pattern](https://en.wikipedia.org/wiki/Facade_pattern).  That said, we're talking about terminology that was published in 1994, so if it's too outdated to be recognizable to a developer today, I'm fine with changing it to something else.

I would like to avoid calling it a context, though.  Having a `GraphQLContext` that comprises a `GlobalContext` and a `RequestContext` is a tolerable level of terminology reuse, because they all have the same purpose -- to convey data from outside the request handling code into the recursive structure of resolver calls.  What was called `ResolverContext` really has a different purpose: simplifying interaction with the rest of the Warpgrapher framework.  The idea that you'd make a call on your context, to get a context, from which you get a context (Resolver, GraphQL, and Global or Request, respectively) is where I had trouble.

src/client/mod.rs
maldrake: Yes, it is possible.  There are a handful of places where the interaction, or lack thereof, between default type parameters and type inference are described, but the one at [https://users.rust-lang.org/t/default-trait-type-not-working-for-type-inference/33905/2](https://users.rust-lang.org/t/default-trait-type-not-working-for-type-inference/33905/2) was the clearest and most succinct that I've found.

However, the syntax for the workaround isn't great, either.  It would be `<Client>::new_with_http` rather than `Client::<(), ()>`.  It's a little more concise, but I suspect most novice and even some experience Rust developers would not have encountered that syntax before.

Another option would be to introduce a trait in the middle somewhere, capturing the type parameters to Engine as associated types.  However, that would force the user to create a struct implementing that trait, specifying the actuals for the associated types, and implementing at minimum a get/set pair for an engine, which seemed even clunkier.

README.md
maldrake: Agreed.  Added.

book/src/warpgrapher/databases.md
maldrake: Agreed.  Added.

src/client/mod.rs
maldrake: Agreed.  Changed.

src/engine/database/cosmos.rs
```diff
@@ -0,0 +1,922 @@
...
+/// build a database connection pool.
+///
+/// # Examples
+///
+/// ```rust,no_run
+/// # use warpgrapher::Error;
+/// # use warpgrapher::engine::database::cosmos::CosmosEndpoint;
+/// #
+/// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+///     let ce = CosmosEndpoint::from_env()?;
```
shellforge, line 29: I just noticed this syntax. What are the `#` inside the example code block?

src/lib.rs
```diff
@@ -1,3 +1,4 @@
+// t1
```
shellforge: This line should go, I forgot to remove it.

src/client/mod.rs
```diff
@@ -1,322 +1,498 @@
...
-
-        // extract data from result
-        match body.get("data") {
-            None => Err(Error::new(
-                ErrorKind::ClientRequestUnexpectedPayload(body.to_owned()),
-                None,
-            )),
-            Some(data) => Ok(data.to_owned()),
-        }
+        body.as_object_mut()
```
shellforge: @maldrake : I assume the idea behind this block is to take a graphql output that looks like:

```json
{
  "data": {
    "Project": <PROJECT>
  }
}
```

And return the user just `<PROJECT>` without the extra two layers of indirection. I think that in most cases (certainly in all the CRUD cases) this is the desired behavior. However, GraphQL supports multiple queries in a single request. For example, the following query:

```
query {
  Project {
    id
  }
  User {
    id
  }
}
```

Would return the following output:

```json
{
  "data": {
    "Project": <PROJECT>,
    "User": <USER>
  }
}
```

And this approach will prevent the caller from taking advantage of that functionality. What would you think of moving the logic that strips out the `result_field` to the individual CRUD functions (i.e. `create_node`) and allow this function to return the entire data object?

src/engine/database/cosmos.rs
maldrake: It hides the line from the documentation output, so you can include lines you need to make the doctest compile that don't have instructive value as part of documentation.

src/lib.rs
maldrake: Agreed.  Will be resolved in the next push.

src/client/mod.rs
maldrake: Good catch.  Will be resolved in the next push.

shellforge (approved):

PR #35, opened 2020-03-31
Async Client

shellforge:


.github/workflows/audit.yml
```diff
@@ -0,0 +1,35 @@
...
+    - name: Run Cargo Audit
+      run: rustbins/bin/cargo-audit audit || echo "Audit failed!"
+        
+    - name: Create security issue if audit fails
+      if: failure()
+      uses: JasonEtco/create-an-issue@v2
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        title: Cargo audit failure
```
maldrake: Will this create a new issue every time a CI/CD build job runs and cargo audit fails?  Or will it de-depulicate, so that if it takes a while for our dependencies to fix the issue, we only get the one issue opened?

src/client/mod.rs
```diff
@@ -18,43 +17,49 @@ use std::collections::BTreeMap;
 /// use std::env::var_os;
 /// use warpgrapher::client::graphql;
 ///
-/// let query = "query { Project { id name } }";
-/// let results = graphql("http://localhost:5000/graphql".to_owned(), query.to_owned(), None);
-/// let projects = results.unwrap().get("Project");
+/// #[tokio::main]
```
maldrake: Should this have a norun tag? Doc tests shouldn't fail because you don't have a server instance up and running on your machine.

src/client/mod.rs
```diff
@@ -18,43 +17,49 @@ use std::collections::BTreeMap;
...
-        .header("Content-Type", "application/json")
-        .send_json(&req_body)
+
+    // send request
+    let client = reqwest::Client::new();
+    let resp = client.post(endpoint.as_str())
+        .json(&req_body)
+        .send()
         .await
         .map_err(|e| Error::new(ErrorKind::ClientRequestFailed(format!("{:#?}", e)), None))?;
```
maldrake: Why format the original error as a string, rather than passing it back whole within the ClientRequestFailed ErrorKind?

src/client/mod.rs
```diff
@@ -18,43 +17,49 @@ use std::collections::BTreeMap;
...
+    
+    // extract data from result
     match body.get("data") {
         None => Err(Error::new(
             ErrorKind::ClientRequestUnexpectedPayload(body.to_owned()),
             None,
         )),
         Some(data) => {
-            debug!("Result Data: {:#?}", data);
+            //debug!("Result Data: {:#?}", data);
```
maldrake: As a matter of cleanliness, I'm not that fond of leaving commented code in the main branch.  I would suggest that if this debug statement isn't needed at all, then delete it entirely rather than commenting it.  If it's useful, but not something helpful to a library client debugging their own code, I'd make it a trace statement.

src/client/mod.rs
```diff
@@ -97,6 +102,37 @@ impl WarpgrapherClient {
...
+    ///
+    ///     let query = "query { Project { id name } }";
+    ///     let results = client.graphql(
+    ///         "query { Project { id name } }".to_string(),
+    ///         None
+    ///     ).await;
+    /// }
+    /// ```
+    #[allow(clippy::needless_doctest_main)]
+    pub async fn graphql(
```
maldrake, line 78: Why provide this as both a client method and a standalone function?  Would it avoid confusion among options to have the WarpgrapherClient be the sole mechanism to make queries?

src/client/mod.rs
```diff
@@ -18,43 +17,49 @@ use std::collections::BTreeMap;
...
-/// let projects = results.unwrap().get("Project");
+/// #[tokio::main]
+/// async fn main() {
+///     let query = "query { Project { id name } }";
+///     let results = graphql("http://localhost:5000/graphql".to_owned(), query.to_owned(), None).await;
+///     let projects = results.unwrap().get("Project");
+/// }
 /// ```
-#[actix_rt::main]
+#[allow(clippy::needless_doctest_main)]
```
maldrake: It is possible to use the #[feature(async_await)] line at the top of the doctest and eliminate the main clause, thus getting rid of the need for the clippy allow directive?

src/server/schema.rs
```diff
@@ -2118,11 +2118,11 @@ pub fn create_root_node<GlobalCtx: Debug, ReqCtx: Debug + WarpgrapherRequestCont
         ))
     })
     .map_err(|e| {
-        (e.downcast::<Error>()
+        e.downcast::<Error>()
             .and_then(|e| Ok(*e))
             .unwrap_or_else(|e| {
                 Error::new(ErrorKind::MissingSchemaElement(format!("{:#?}", e)), None)
```
maldrake, line 2124: What is the String in this error actually supposed to be?  The name of the missing element that was requested?  Or something else?  Is it really supposed to be a debug string from a source error?  Why not include the error directly in the ErrorKind, rather than a string dump of it?

src/client/mod.rs
```diff
@@ -18,43 +17,49 @@ use std::collections::BTreeMap;
...
 pub async fn graphql(
     endpoint: String,
```
maldrake: Why a String and not an &str?

src/client/mod.rs
```diff
@@ -18,43 +17,49 @@ use std::collections::BTreeMap;
...
     query: String,
```
maldrake: Why a String and not an &str?

src/client/mod.rs
```diff
@@ -18,43 +17,49 @@ use std::collections::BTreeMap;
...
     input: Option<Value>,
```
maldrake: Is ownership necessary, or would a reference suffice?  Ownership is generally better than taking a reference and cloning, but if only a reference is needed, that's even more flexible.

src/client/mod.rs
```diff
@@ -105,20 +141,24 @@ impl WarpgrapherClient {
...
-    /// ```rust,no_run
+    /// ```no_run
     /// use serde_json::json;
     /// use std::env::var_os;
     /// use warpgrapher::WarpgrapherClient;
     ///
-    /// let mut client = WarpgrapherClient::new("http://localhost:5000/graphql");
+    /// #[tokio::main]
+    /// async fn main() {
+    ///     let mut client = WarpgrapherClient::new("http://localhost:5000/graphql");
```
maldrake, line 138: WarpgrapherClient could probably eagerly implement a few more traits from the set of Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Display, and Default.  Some of them may not make sense in this context, but some probably do.

src/client/mod.rs
```diff
@@ -18,43 +17,49 @@ use std::collections::BTreeMap;
...
 /// use warpgrapher::client::graphql;
 ///
-/// let query = "query { Project { id name } }";
-/// let results = graphql("http://localhost:5000/graphql".to_owned(), query.to_owned(), None);
-/// let projects = results.unwrap().get("Project");
+/// #[tokio::main]
+/// async fn main() {
+///     let query = "query { Project { id name } }";
+///     let results = graphql("http://localhost:5000/graphql".to_owned(), query.to_owned(), None).await;
+///     let projects = results.unwrap().get("Project");
```
maldrake: Ideally, doc examples should not use unwrap.

src/client/mod.rs
```diff
@@ -18,43 +17,49 @@ use std::collections::BTreeMap;
...
+/// }
 /// ```
-#[actix_rt::main]
+#[allow(clippy::needless_doctest_main)]
 pub async fn graphql(
     endpoint: String,
     query: String,
     input: Option<Value>,
 ) -> Result<Value, Error> {
```
maldrake: The documentation comment should describe the conditions under which an error is returned.

src/client/mod.rs
```diff
@@ -97,6 +102,37 @@ impl WarpgrapherClient {
             endpoint: endpoint.to_string(),
         }
     }
+    
+    /// Executes a raw graphql query.
+    ///
```
maldrake, line 49: The documentation should describe conditions under which error(s) are returned.

src/client/mod.rs
```diff
@@ -97,6 +102,37 @@ impl WarpgrapherClient {
...
+    pub async fn graphql(
+        &mut self,
+        query: String,
+        input: Option<Value>
+    ) -> Result<Value, Error> {
+        graphql(self.endpoint.clone(), query, input).await
+    }
 
     /// Takes the name of a WarpgrapherType and executes a NodeCreate operation. Requires
     /// a query shape and the input of the node being created.
```
maldrake, line 118: The documentation should describe conditions under which error(s) are returned.

src/client/mod.rs
```diff
@@ -142,23 +182,27 @@ impl WarpgrapherClient {
     ///
```
maldrake, line 169: The documentation should describe conditions under which error(s) are returned.

src/client/mod.rs
```diff
@@ -191,19 +235,24 @@ impl WarpgrapherClient {
...
```
maldrake, line 229: The documentation should describe conditions under which error(s) are returned.

src/client/mod.rs
```diff
@@ -230,20 +279,24 @@ impl WarpgrapherClient {
...
```
maldrake, line 280: The documentation should describe conditions under which errors are returned.

src/client/mod.rs
```diff
@@ -281,15 +334,19 @@ impl WarpgrapherClient {
...
```
maldrake, line 348: The documentation should describe conditions under which errors are returned.

src/client/mod.rs
```diff
@@ -317,14 +374,18 @@ impl WarpgrapherClient {
     /// use std::env::var_os;
```
maldrake, line 402: The documentation (a little above), should describe conditions under which errors are returned.

src/client/mod.rs
```diff
@@ -356,16 +417,20 @@ impl WarpgrapherClient {
...
```
maldrake, line 452: The documentation just above should describe conditions under which errors are returned.

src/client/mod.rs
```diff
@@ -397,15 +462,19 @@ impl WarpgrapherClient {
...
```
maldrake, line 504: The documentation just above should include a description of conditions under which errors are returned.

src/client/mod.rs
```diff
@@ -300,7 +357,7 @@ impl WarpgrapherClient {
             "WarpGrapherClient read_node -- query: {:#?}, input: {:#?}",
             query, input
         );
-        let result = graphql(self.endpoint.to_owned(), query, input.cloned())?;
+        let result = graphql(self.endpoint.to_owned(), query, input.cloned()).await?;
```
maldrake: If ownership is needed here, better to take ownership in the argument, above, rather than taking a reference.  Let the caller decide whether to clone.

src/client/mod.rs
```diff
@@ -336,7 +397,7 @@ impl WarpgrapherClient {
             "WarpGrapherClient read_rel -- query: {:#?}, input: {:#?}",
             query, input
         );
-        let result = graphql(self.endpoint.to_owned(), query, input.cloned())?;
+        let result = graphql(self.endpoint.to_owned(), query, input.cloned()).await?;
```
maldrake: If ownership is needed here, consider making the function signature take ownership and let the caller control cloning.

src/client/mod.rs
shellforge: Doesn't it? On line 16?

src/server/schema.rs
shellforge: This wasn't part of the PR except that clippy was yelling about the extra parenthesis.

src/client/mod.rs
maldrake: Why, yes.  It does.  Not sure how I missed it, as it's quite glaring.  My illiteracy knows no bounds.

src/server/schema.rs
maldrake: Good process question, then.  Is it fair game to fix something we didn't notice in a previous review, if we notice it in a subsequent one?  Or would you rather handle that with a separate issue?

shellforge: I'm game for fixing issues wherever we spot them. In this particular case I don't think I wrote the original code so I didn't have an immediate answer so I was planning on opening a separate ticket to look into afterwards.

src/client/mod.rs
shellforge: Does "eagerly implement" refer to using the derive feature?

maldrake: Yes.  Sorry, by eagerly implement, I just mean defaulting to deriving those traits unless it doesn't make sense to do so.

.github/workflows/audit.yml
shellforge: It turns out that this github job as written will not have the intended effect. Short answer is it will not deduplicate. I tried to account for this by only running this job when a library changes (by looking at Cargo.lock), but it is now obvious to me that that approach will no work in situations where a new vuln is reported. I'm going to remove this job from this PR and add it to another ticket.

src/client/mod.rs
shellforge: I did not have luck with this. The following

```
    /// # Examples
    ///
    /// ```rust,no_run
    /// #[feature(async_await)] 
    /// use serde_json::json;
    /// use std::env::var_os;
    /// use warpgrapher::WarpgrapherClient;
    ///
    ///     let mut client = WarpgrapherClient::new("http://localhost:5000/graphql");
    ///
    ///     let query = "query { Project { id name } }";
    ///     let results = client.graphql(
    ///         "query { Project { id name } }",
    ///         None
    ///     ).await;
    /// ```
```

yields

```
failures:

---- src/client/mod.rs - client::WarpgrapherClient::graphql (line 61) stdout ----
error[E0728]: `await` is only allowed inside `async` functions and blocks
  --> src/client/mod.rs:70:19
   |
3  |   fn main() {
   |      ---- this is not `async`
...
12 |       let results = client.graphql(
   |  ___________________^
13 | |         "query { Project { id name } }",
14 | |         None
15 | |     ).await;
   | |___________^ only allowed inside `async` functions and blocks
```

Have you had this approach work?

maldrake: I could try to explain the convoluted mistakes of both reasoning and test cases that led me to suggest this, but the tl;dr version is that my (corrected) results now match yours.  This is not a thing, and so please disregard this comment thread entirely.

maldrake (approved):

PR #38, opened 2020-04-21
Exports Cleanup

shellforge:


examples/project-tracker/actix_server.rs
```diff
@@ -11,16 +11,16 @@ use actix_web::{web, App, HttpResponse, HttpServer};
 use juniper::http::GraphQLRequest;
 use std::collections::HashMap;
 
-use super::{GlobalContext, ReqContext};
-use warpgrapher::Engine;
+use super::{AppGlobalContext, AppRequestContext};
```
maldrake, line 14: This would probably be handled under a separate issue, but what would you think about pulling the project-tracker example out to a separate repo altogether.  Now that we've coverted from server to engine, this seems to pull in a lot of irrelevant dependencies.

src/lib.rs
```diff
@@ -10,23 +10,9 @@
...
-pub use engine::config::{
-    WarpgrapherConfig, WarpgrapherEndpoint, WarpgrapherResolverFunc, WarpgrapherResolvers,
-    WarpgrapherType, WarpgrapherValidatorFunc, WarpgrapherValidators,
-};
-pub use engine::context::{GraphQLContext, WarpgrapherRequestContext};
-pub use engine::extensions::{Extension, WarpgrapherExtensions};
-pub use engine::neo4j::Neo4jEndpoint;
-pub use engine::objects::Node;
-pub use engine::schema::{Info, Property};
-pub use engine::Engine;
```
maldrake, line 28: I'd be okay with re-exporting engine::config::Config and engine::Engine, given that if I remember correctly, every client that uses Warpgrapher has to use those two types.  Other than that, the drastic cutting down of re-exports looks good.  (And I could be talked out of the two above, if you'd rather not.)

shellforge: I agree with that suggestion.

examples/project-tracker/actix_server.rs
shellforge: Agreed. I'll open up a ticket for this.

shellforge: https://github.com/warpforge/warpgrapher/issues/41

maldrake (approved):

PR #40, opened 2020-04-23
Improve custom resolver handling

shellforge:


shellforge:
Rename convenience functions to "resolve" instead of "return".

maldrake (approved):

PR #43, opened 2020-05-04
Query Fragments bugfix

shellforge:


src/engine/objects.rs
```diff
@@ -465,6 +467,17 @@ where
             self.concrete_typename
         );
 
+        // this mismatch can occur when query fragments are used. correct
```
shellforge, line 470: @maldrake : this is the only real change in this bugfix. All the other diffs are from cargo fmt.

maldrake (approved):

PR #46, opened 2020-05-14
Updated README links and book html

shellforge:


docs/api/crud_nodes.md.old
```diff
@@ -0,0 +1,275 @@
+# Node CRUD Operations
```
maldrake: Why check in a file with a .old extension?

docs/api/crud_rels.md.old
```diff
@@ -0,0 +1,162 @@
+# Relationship CRUD Operations
```
maldrake: Why check in a file with a .old extension?

docs/api/crud_nodes.md.old
shellforge: It's more fun that way.

maldrake (approved):

PR #47, opened 2020-05-15
Removed Cargo.lock

shellforge:
Still having issues publishing the crate:

```
Run cargo publish --token ***
  cargo publish --token ***
  shell: /bin/bash -e {0}
    Updating crates.io index
error: 1 files in the working directory contain changes that were not yet committed into git:

Cargo.lock
```

In the course of looking for the source of the issue, ran across the cargo book which states that `Cargo.lock` should not be committed in the repo (https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries).

Hoping that by following best practices the issues will resolve itself.


shellforge:
@maldrake : would like your thoughts on this PR.

maldrake (approved):
I never really understood why we'd been committing it.  I'm fine with removing it from the repo.

PR #52, opened 2020-07-22
Graphson Support Updates

shellforge:


src/engine/config.rs
```diff
@@ -1129,6 +1129,36 @@ impl Type {
...
+    /// # use warpgrapher::engine::config::{Type};
+    ///
+    /// let t = Type::from_yaml("
+    /// name: User
+    /// props:
+    ///   - name: name
+    ///     type: String
+    /// ").unwrap();
+    /// ```
+    pub fn from_yaml(yaml: &str) -> Result<Type, Error> {
```
maldrake: Given that this is the only way to get a type from a &str, would it make sense to make this a TryFrom<&str> implementation instead?

src/engine/database/neo4j.rs
```diff
@@ -30,15 +30,20 @@ use tokio::runtime::Runtime;
...
+///     };
 /// #    Ok(())
 /// # }
 /// ```
 pub struct Neo4jEndpoint {
-    host: String,
-    port: u16,
-    user: String,
-    pass: String,
+    pub host: String,
```
maldrake: We may have talked about this, so apologies if I've forgotten, but what was the use case / rationale for making these public?

src/engine/database/neo4j.rs
```diff
@@ -805,3 +810,19 @@ impl From<Value> for bolt_proto::value::Value {
         }
     }
 }
+
+// TODO: fix this (jeffrey: could use your advise)
```
maldrake: You're definitely up against the limit of not being able to define a trait implementation where both the trait and struct are imported from other crates.  One option is the way you've done it, though I would consider renaming things, as you're not really converting to a Warpgrapher Value -- you're converting to a HashMap from String to Warpgraher Values.

Your other option would be to `impl TryFrom<HashMap<String, bolt_proto::value::Value>> for Value` and return a Value::Map, from which you could then borrow or extract the HashMap<String, Value> as needed.  This latter option is probably what I'd go for.

src/engine/objects/mod.rs
```diff
@@ -177,7 +177,19 @@ where
...
+    /// ```rust
+    /// use std::collections::HashMap;
+    /// use warpgrapher::engine::objects::Node;
+    /// use warpgrapher::engine::value::Value;
+    ///
+    /// let mut props = HashMap::<String, Value>::new();
+    /// props.insert("name".to_string(), "Joe".to_string().into());
+    /// let user_node = Node::<(),()>::new("User".to_string(), props);
+    /// ```
+    pub fn new(
```
maldrake: Does Node::new really need to be made public? Aren't custom resolvers going through the ResolverFacade to create nodes?

src/engine/objects/mod.rs
```diff
@@ -586,8 +612,10 @@ where
     }
 }
 
+/// Represents a reference to a [`Node`] object as either an [`Identifier`]
+/// containing a type and id, or a complete [`Node`] struct.
 #[derive(Clone, Debug)]
-pub(crate) enum NodeRef<GlobalCtx: GlobalContext, RequestCtx: RequestContext> {
+pub enum NodeRef<GlobalCtx: GlobalContext, RequestCtx: RequestContext> {
```
maldrake: Do custom resolvers need to create these directly, as opposed to going through the ResolverFacade?

src/engine/objects/mod.rs
```diff
@@ -648,7 +676,7 @@ where
     GlobalCtx: GlobalContext,
     RequestCtx: RequestContext,
 {
-    pub(crate) fn new(
+    pub fn new(
```
maldrake: Do custom resolvers need to create the directly, rather than going through the ResolverFacade?

src/engine/resolvers.rs
```diff
@@ -80,11 +82,60 @@ where
...
+    ///
+    ///         let neo4j_client = facade.db_as_neo4j().await.unwrap();
+    ///         // use neo4j client
+    ///
+    ///     });
+    ///     facade.resolve_null()
+    /// }
+    /// ```
+    #[cfg(feature = "neo4j")]
+    pub async fn db_as_neo4j(
```
maldrake: Would it make sense to put this method on the DatabasePool itself?  So the call would be `resolver.pool().as_neo4j()`. Pool just pulls the databasepool out of self.executor.context().  It's up to the DatabasePool to know how to deliver itself as a neo4j client.

Also, from this point forward, we should maintain feature parity between Cosmos and Neo4J.  If we need this convenience method for neo4j, shouldn't we also have it for Cosmos?

src/engine/resolvers.rs
```diff
@@ -80,11 +82,60 @@ where
...
+        &self,
+    ) -> Result<bb8::PooledConnection<'_, bb8_bolt::BoltConnectionManager>, Error> {
+        let db_pool: &bb8::Pool<bb8_bolt::BoltConnectionManager> =
+            match self.executor().context().pool() {
+                DatabasePool::Neo4j(p) => p,
+                _ => {
+                    return Err(Error::DatabaseMismatch {});
```
maldrake: Do we need a new error, or can we make Error::DatabaseNotFound broad enough to cover this case. It's already everywhere else that we look look for a database of a particular kind and can't find one.  If we really do need this to be a separate error, it should follow the NounVerb naming pattern of the rest of the errors, not NounNoun.

src/engine/resolvers.rs
```diff
@@ -80,11 +82,60 @@ where
...
+                }
+            };
+        let neo4j_client = db_pool
+            .get()
```
maldrake: Given this call, this may not be appropriate for the naming convention of as_neo4j().  It's probably more appropriate for it to be into_neo4j(), given that it's making a call to a call with variable expense to a non-copy operation in order to produce the result.

src/engine/value.rs
```diff
@@ -26,6 +26,33 @@ pub enum Value {
     UInt64(u64),
 }
 
+impl Value {
```
maldrake: Why not implement this as `impl TryFrom<Value> for String`?

src/error.rs
```diff
@@ -59,6 +59,10 @@ pub enum Error {
     /// never be done in production
     DatabaseNotFound,
 
+    /// Returned if attempting to access a database driver for a different database type
+    /// than the one configured.
+    DatabaseMismatch,
```
maldrake: I think it may be possible, and definitely beneficial if we can, to combine this with DatabaseNotFound, rather than adding a new Error variant.  If we really do need to add an error variant, it should follow the same NounVerb naming convention as the others, not NounNoun.

src/error.rs
```diff
@@ -207,6 +217,9 @@ impl Display for Error {
             Error::DatabaseNotFound => {
                 write!(f, "Use of resolvers required a database back-end. Please select either cosmos or neo4j.")
             }
+            Error::DatabaseMismatch => {
+                write!(f, "The database connection you asked for does not match the configured database back-end.")
```
maldrake: It's a nit-picky stylistic thing, but I grew up with the idea that technical documentation should avoid second person.  (Yeah, I know... it appears in the doc comments, especially in the error documentation.)  Up to you... what's our style guide here?

src/engine/config.rs
shellforge: I would say so.

src/engine/database/neo4j.rs
shellforge: Need to instantiate an database endpoint from a source other than environment variables.

src/engine/objects/mod.rs
shellforge: Yes, that was on oversight.

src/engine/objects/mod.rs
shellforge: I believe so, likely also an oversight.

src/engine/database/neo4j.rs
maldrake: Why does that require making the fields of the struct public?  Wouldn't that work as long as you have a `Neo4JEndpoint::new` function?

src/engine/resolvers.rs
shellforge: If you are ok with it, I'm good to reuse "DatabaseNotFound".

src/engine/resolvers.rs
shellforge: No objection.

src/engine/resolvers.rs
maldrake: I'm good with reusing it.  The function caller is asking for a Neo4J database client.  A Neo4J database client is not found.

src/engine/value.rs
shellforge: I think I tried that and I couldn't get it to work, probably because I was doing something wrong. I'll go back and try it again.

src/error.rs
shellforge: Agreed.

src/error.rs
shellforge: That didn't even occur to me. I have no objections to avoiding second person.

src/engine/config.rs
shellforge: Done.

src/engine/database/neo4j.rs
shellforge: Done.

src/engine/config.rs
shellforge: Done.

src/engine/database/neo4j.rs
shellforge: Fixed.

src/engine/database/neo4j.rs
shellforge: Implemented the latter suggestion.

src/engine/objects/mod.rs
shellforge: Fixed.

src/engine/objects/mod.rs
shellforge: Fixed.

src/engine/resolvers.rs
shellforge: Fixed.

src/engine/resolvers.rs
shellforge: Fixed.

src/engine/value.rs
shellforge: Fixed.

src/error.rs
shellforge: Fixed.

src/error.rs
shellforge: Fixed.

src/engine/database/neo4j.rs
```diff
@@ -805,3 +837,56 @@ impl From<Value> for bolt_proto::value::Value {
...
+                        return Err(Error::TypeConversionFailed {
+                            src: "Value".to_string(),
+                            dst: "HashMap::<String, Value>".to_string(),
+                        })
+                    }
+                };
+                Ok(Node::new(type_name.to_string(), props))
+            }
+            _ => Err(Error::TypeConversionFailed {
+                src: "bolt_proto::value::Value".to_string(),
```
maldrake: The documentation comment for the `Error::TypeConversionFailed` error type notes that the `src` string can contain either the src type name or the src value itself that failed to convert. In this case, I think it'd be better to have `src: format!("{:#?}", value),`. My rationale is that the bolt_proto::value::Value enumeration can be any of a number of different variants. Providing the contents of the value will make it easier for a client to debug what went wrong, as they'll see what variant was actually contained in `value`, where they expected a node, which may be valuable debugging information.

src/engine/database/neo4j.rs
```diff
@@ -805,3 +837,56 @@ impl From<Value> for bolt_proto::value::Value {
...
+    RequestCtx: RequestContext,
+{
+    type Error = crate::Error;
+
+    fn try_from(value: bolt_proto::value::Value) -> Result<Self, Error> {
+        match value {
+            bolt_proto::value::Value::Node(n) => {
+                let type_name = &n.labels()[0];
+                let properties: &HashMap<String, bolt_proto::value::Value> = &n.properties();
+                let props_value = Value::try_from(properties.clone())?;
```
maldrake, line 853: Unfortunately, this clone of the properties truly is necessary, because of the API offered by bolt_client. We should consider looking at submitting a PR to the bolt_client crate to make it possible to deconstruct a node value and take out the properties without cloning them.  If someone is reading a node that contains a large number of properties and/or a number of large properties, block copying the entire data set retrieved from the database could be wasted overhead.

src/engine/database/neo4j.rs
```diff
@@ -805,3 +837,56 @@ impl From<Value> for bolt_proto::value::Value {
...
+    }
+}
+
+impl TryFrom<HashMap<String, bolt_proto::value::Value>> for Value {
+    type Error = Error;
+
+    fn try_from(hm: HashMap<String, bolt_proto::value::Value>) -> Result<Value, Error> {
+        let mut hmv = HashMap::<String, Value>::new();
+        for (key, bolt_value) in hm.iter() {
+            let value = match Value::try_from(bolt_value.clone()) {
```
maldrake: Is this clone operation necessary?  You're getting ownership of the `hm` `HashMap` in the parameters, so you should be able to use `hm.into_iter()` in the loop construct above, taking ownership instead of just getting a reference. Then you could call `Value::try_from` on the `bolt_value` directly, avoiding the unnecessary clone of the data value.

src/engine/database/neo4j.rs
```diff
@@ -805,3 +837,56 @@ impl From<Value> for bolt_proto::value::Value {
...
+        }
+    }
+}
+
+impl TryFrom<HashMap<String, bolt_proto::value::Value>> for Value {
+    type Error = Error;
+
+    fn try_from(hm: HashMap<String, bolt_proto::value::Value>) -> Result<Value, Error> {
+        let mut hmv = HashMap::<String, Value>::new();
+        for (key, bolt_value) in hm.iter() {
```
maldrake: We seem to have adopted a more functional style in Warpgrapher, since figuring out it works just fine.  Instead of a for loop, this could be a map/collect operation over the iterator, rather than a for loop.

src/engine/database/neo4j.rs
```diff
@@ -805,3 +837,56 @@ impl From<Value> for bolt_proto::value::Value {
...
+            let value = match Value::try_from(bolt_value.clone()) {
+                Ok(v) => v,
+                Err(_) => {
+                    return Err(Error::TypeConversionFailed {
+                        src: "Value".to_string(),
```
maldrake: Because `value` is an enumerated type with multiple variants, rather than a primitive type, I would format the debugging string into src, rather than just the type name.

src/engine/objects/mod.rs
```diff
@@ -189,6 +189,20 @@ where
         }
     }
 
+    /// Returns a new Node.
```
maldrake: Copy/paste error?  The documentation comment says this function returns a `Node`, but it seems to return fields.

src/engine/resolvers.rs
```diff
@@ -3,14 +3,16 @@
 
 use crate::engine::context::GraphQLContext;
 use crate::engine::context::{GlobalContext, RequestContext};
+#[cfg(feature = "neo4j")]
```
maldrake: This seems odd to me.  I would think we wouldn't have a neo4j crate feature guard here if we had feature parity among all database types.

src/engine/resolvers.rs
```diff
@@ -80,11 +82,60 @@ where
...
+    ///
+    ///         let neo4j_client = facade.db_into_neo4j().await.unwrap();
+    ///         // use neo4j client
+    ///
+    ///     });
+    ///     facade.resolve_null()
+    /// }
+    /// ```
+    #[cfg(feature = "neo4j")]
+    pub async fn db_into_neo4j(
```
maldrake, line 112: Unless overruled, I still think it makes more sense for this into operation to be on the pool itself, so that the call chain for a client would look like `facade.pool().into_neo4j() ` rather than `facade.db_into_neo4j()`.  Or given that this is a facade that's supposed to make life easier, I think I'd be fine with a `facade.db_into_neo4j()` call, as the interface is here, but with the behavior delegated under the hood so that the code in this function is just a one-line `self.executor().pool().into_neo4j()`.  The functionality for doing the conversion into a Neo4J client should live on the pool, though, not in the facade, I think.

maldrake, line 112: To maintain feature parity with Cosmos, there should be a db_into_cosmos function as well.

src/engine/value.rs
```diff
@@ -261,6 +261,20 @@ where
...
 }
 
+impl TryFrom<Value> for HashMap<String, Value> {
+    type Error = Error;
+
+    fn try_from(value: Value) -> Result<HashMap<String, Value>, Error> {
+        match value {
+            Value::Map(hm) => Ok(hm),
+            _ => Err(Error::TypeConversionFailed {
+                src: "Value".to_string(),
```
maldrake: I would recommend putting a formatted debug output of the `value` here, as opposed to just the Value type name, as it will provide richer information for debugging.

src/engine/resolvers.rs
```diff
@@ -80,11 +82,60 @@ where
...
+            match self.executor().context().pool() {
+                DatabasePool::Neo4j(p) => p,
+                _ => {
+                    return Err(Error::DatabaseNotFound {});
+                }
+            };
+        let neo4j_client = db_pool
+            .get()
+            .await
+            .map_err(|e| Error::Neo4jPoolGetConnectionFailed { source: e })?;
```
maldrake: With the From interface recommended in a comment below, it should be possible to simplify this code, as the error type can be more easily/automagically converted.

src/error.rs
```diff
@@ -89,6 +89,12 @@ pub enum Error {
         message: bolt_proto::message::Message,
     },
 
+    // Returned if a bb8 connection cannnot be obtained form the pool
+    #[cfg(feature = "neo4j")]
+    Neo4jPoolGetConnectionFailed {
+        source: bb8::RunError<bb8_bolt::Error>,
```
maldrake, line 95: I would recommend also adding an `impl From<bb8:RunError<bb8_bolt::Error>> for Error` block below, for convenience and consistency with other error handling.

src/engine/database/neo4j.rs
```diff
@@ -42,6 +47,33 @@ pub struct Neo4jEndpoint {
...
+    ///     let ne = Neo4jEndpoint::new(
+    ///         "127.0.0.1".to_string(),
+    ///         7687,
+    ///         "neo4j".to_string(),
+    ///         "password".to_string()
+    ///     );
+    /// #    Ok(())
+    /// # }
+    /// ```
+    pub fn new(host: String, port: u16, user: String, pass: String) -> Self {
```
maldrake, line 68: Stylistic question... in most cases, we've made these kinds of string configuration parameters &str types, so that it's possible to pass in strings like "127.0.0.1" without the caller having to call .to_string() on a number of arguments.  The downside is, of course, the overhead of an extra string copy, if someone would've actually passed in a real `String`. But they're typically short values, not many-kilobyte text files, so it's not that much overhead.... especially for something that's not in a critical loop -- one typically only sets up a database endpoint once.  Any particular reason you made the arguments `String` instead of `&str`?

(I know some people go all the way to parameterizing these kinds of arguments as Into<String>, that I've not been a fan of that, as it clutters up the type signature of the function call too much for my liking.)

src/engine/resolvers.rs
```diff
@@ -80,11 +82,60 @@ where
...
+    /// # use tokio::runtime::Runtime;
+    /// # use warpgrapher::engine::resolvers::{ResolverFacade, ExecutionResult};
+    ///
+    /// fn custom_resolve(facade: ResolverFacade<(), ()>) -> ExecutionResult {
+    ///     let mut rt = Runtime::new().unwrap();
+    ///     rt.block_on(async {
+    ///
+    ///         let neo4j_client = facade.db_into_neo4j().await.unwrap();
+    ///         // use neo4j client
+    ///
```
maldrake, line 106: Education moment for me... is the client not safe to pass back out of the async block, such that we could hide the rt.block_on call inside the db_into_neo4j function and not burden the client with it?

src/engine/database/neo4j.rs
shellforge: I originally had a fold operation on the iterator, but there does not appear to be a clean way to propagate an error up from a closure in a fold (and since we are performing a try_from, we need to). Do you have any ideas to work around that? (I do prefer a functional style, I just couldn't make it work). 

That being said I would lean against adopting just a single style in the project. Performance and usability always come first, but all other things equal, I personally would defer to individual contributor on style.

maldrake: Ah, there is a solution for that.  Iterators have a try_fold method.

That said, I can take the hint on not going overboard with stylistic comments -- I definitely need to get better at distinguishing between what's a genuine readability / maintainability benefit and what's just personal preference of little or no substance and consequence.

src/engine/database/neo4j.rs
shellforge: I was operating under the assumption that the arguments for this function in particular will be derived from some source in real time (config file) and will likely be strings. Providing something like a password as a `&str` I suspect with would be the exception rather than the norm, and in those few cases calling `.to_string()` would not be a major inconvenience. 

I'm open to alternatives, however. Would you be in favor on switching the inputs to `&str` and then calling `.to_string()` on each of them in the constructor (since the fields in the struct are `String`s. ?

src/engine/resolvers.rs
shellforge: I suspect this is going to lead to a deeper discussion. Let's chat about this one.

src/engine/database/neo4j.rs
maldrake: Good point.  I'm good with `String`s.

src/engine/database/neo4j.rs
shellforge: I could use some help here. If I try `src: format!("{:#?}", bolt_value)`, I get a "value borrowed after move". My naive fix is to clone `bolt_value` when I pass it to `Value::try_from` above, but we are trying to avoid cloning. Any ideas?

src/engine/database/neo4j.rs
shellforge: Fixed.

src/engine/database/neo4j.rs
shellforge: Agreed, let's look into this.

src/engine/database/neo4j.rs
maldrake: Hmmm... what about `format!("{:#?}", &bolt_value)`?  If I recall correctly, the `format!` macro should be able to work off a reference. I don't think it needs ownership.

src/engine/database/neo4j.rs
shellforge: try_fold is what I was looking for! Fixed.

src/engine/value.rs
shellforge: Fixed.

src/engine/resolvers.rs
shellforge: Fixed.

src/error.rs
shellforge: Fixed.

src/engine/database/neo4j.rs
shellforge: Using `if let` still doesn't fix the move after borrow issue.

src/engine/objects/mod.rs
shellforge: Fixed.

src/engine/database/neo4j.rs
maldrake: That confuses me.  Maybe we can set up a screen share at your convenience?

src/engine/resolvers.rs
shellforge: Fixed.

src/engine/resolvers.rs
shellforge: Fixed.

src/engine/resolvers.rs
shellforge: Fixed.

shellforge: Fixed.

src/engine/database/mod.rs
```diff
@@ -52,6 +52,24 @@ pub enum DatabasePool {
     NoDatabase,
 }
 
+impl DatabasePool {
+    #[cfg(feature = "neo4j")]
+    pub(crate) fn neo4j(&self) -> Result<&bb8::Pool<bb8_bolt::BoltConnectionManager>, Error> {
```
maldrake, line 57: I changed these because I realize this isn't really an `into` kind of function.  It's not a type conversion... just getting a connection from a pool.  But, I already commented asking for one change here, and I didn't want to do a "no, the couch really looked better where it was," routine, so I fixed and pushed this one.

src/engine/database/neo4j.rs
```diff
@@ -805,3 +837,57 @@ impl From<Value> for bolt_proto::value::Value {
...
+{
+    type Error = crate::Error;
+
+    fn try_from(value: bolt_proto::value::Value) -> Result<Self, Error> {
+        match value {
+            bolt_proto::value::Value::Node(n) => {
+                let type_name = &n.labels()[0];
+                let properties: &HashMap<String, bolt_proto::value::Value> = &n.properties();
+                let props_value = Value::try_from(properties.clone())?;
+                let props = match HashMap::<String, Value>::try_from(props_value.clone()) {
```
maldrake: Would it be possible to get rid of the props_value.clone() here.  I know we have to clone it out of the node properties from bolt_proto's Value type, but after that, it would be great to avoid copies where we can.  Would it work to do an if let and use the reference in `format!` the way you did below?

src/engine/value.rs
```diff
@@ -261,6 +266,46 @@ where
...
+            Value::Map(hm) => Ok(hm),
+            _ => Err(Error::TypeConversionFailed {
+                src: format!("{:#?}", value),
+                dst: "HashMap<String, Value>".to_string(),
+            }),
+        }
+    }
+}
+
+/*
```
maldrake: In general, I try to avoid commented out code going into master branches.  Is there a reason to keep this, or should it be deleted?

shellforge: That was an oversight. Fixed.

maldrake (approved):

PR #61, opened 2020-08-28
Transactions v2

maldrake:
Refactors resolvers, visitors, and database code to move mutations into a single query, making them transactions across a broader selection of databases.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.freeze
+*.log
 *.swo
 *.swp
 .old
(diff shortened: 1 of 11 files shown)


shellforge:
@maldrake: I reviewed the PR and mechanically I didn't find any issues. When you have a moment, however, I would like to run through some parts together to make sure I understand the high-level/conceptual changes that are taking place (since it's a little tougher to deduce just from looking at diffs).

shellforge (approved):

PR #66, opened 2020-09-07
Added Tinkerpop support.

### LINKED ISSUE #62: Add support for Gremlin v3 Databases
Examples include Apache Tinkerpop and AWS Neptune.

maldrake:
Closes #62.


src/engine/database/mod.rs
```diff
@@ -305,7 +324,7 @@ pub(crate) struct NodeQueryVar {
 }
 
 impl NodeQueryVar {
-    #[cfg(any(feature = "cosmos", feature = "neo4j"))]
+    #[cfg(any(feature = "cosmos", feature = "gremlin", feature = "neo4j"))]
```
shellforge, line 327: If this block is enabled for all features, why is it needed?

tests/fixtures/gremlin-console/remote.yaml
```diff
@@ -0,0 +1,9 @@
+hosts: [127.0.0.1]
+username: stephen
```
shellforge, line 2: Is there a special significance to this username?

tests/fixtures/gremlin-console/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM tinkerpop/gremlin-console:latest
```
shellforge, line 1: Just curious - why did you choose gremlin-console over gremlin-sever for the test fixture?

.github/workflows/test.yml
```diff
@@ -17,14 +17,20 @@ jobs:
     runs-on: ubuntu-latest
 
     env:
-      WG_NEO4J_HOST: 127.0.0.1
-      WG_NEO4J_PORT: 7687
-      WG_NEO4J_USER: neo4j
-      WG_NEO4J_PASS: testpass
       WG_COSMOS_HOST: ${{ secrets.WG_COSMOS_HOST }}
```
shellforge, line 20: Now that we have a local gremlin-based database for testing, can we remove the cosmos test db?

src/engine/database/mod.rs
maldrake: Cargo doesn't allow for features to be a choice of one (or more) of a set.  Each one can be toggled independently, so we must also account for the possibility that all features are false.  In this case, the line that follows the guard would break the build if all features are false, corresponding to a selection of no database at all.

tests/fixtures/gremlin-console/remote.yaml
maldrake: Yes, the Tinkerpop graph database comes with a few default users, and it seemed easier to use their pre-provisioned user graph than manually building our own.

tests/fixtures/gremlin-console/Dockerfile
maldrake: Ah, that's probably a bit of misleading organization. The test fixture is actually the peer directory with the tinkerpop server. The gremlin-console Dockerfile and supporting files aren't actually test fixtures.  Per the text in the Contributing.md file, they're there to give people a way to manually connect to the graph database and run ad hoc queries to investigate system behavior and debug during development, the way you'd use a browser manually connected to Neo4J to debug system interactions with a Neo4J database.  I welcome suggestions on a better place to locate it. It might be better to place it as a root subdirectory to the project rather than in test fixtures, given that it's not really a fixture.  Good catch, and any preferences?

.github/workflows/test.yml
maldrake: I would suggest we keep Cosmos, as it uses some small variations in code paths within Warpgrapher, and very wildly different code paths within the gremlin-client library, because Microsoft is using two different outdated versions of the GraphSON wire format.  I'd suggest removing it if/when Microsoft ups their supported version of GraphSON to v3, which the rest of the world is using now.

tests/fixtures/gremlin-console/Dockerfile
shellforge: Makes sense - no need to change it.

src/engine/database/mod.rs
shellforge: What would cause the build to break if all features were false (i.e. no database selected)? That struct and the impl block don't seem to contain any database-specific variables or logic.

maldrake: In this case, it wouldn't be a hard build break (error), but rather a warning of dead code.

shellforge:
@maldrake: just one final piece - you need to "compile" the book and commit the changes (github looks at the html output instead of the markdown source). There is a script that should do it: https://github.com/warpforge/warpgrapher/blob/add-neptune-support/book/build.sh.

shellforge (approved):

PR #70, opened 2020-09-11
Closes #59. Closes #60.

### LINKED ISSUE #59: Dynamic prop resolvers throw runtime error
```
thread 'main' panicked at 'Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.', /home/dev/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-0.2.22/src/runtime/enter.rs:38:5
```

### LINKED ISSUE #60: Relationship with `_` in name results in panics when building engine
Config:

```
version: 1
model: 
 - name: User
   props:
    - name: name
      type: String
 - name: Project
   props: 
    - name: name
      type: String 
   rels:
     - name: top_contributor
       nodes: [User]
       resolver: resolve_project_top_contributor
```

Error:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: NameParseError("Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"ProjectTop ContributorRel\" does not")',
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'Failed to build engine: SchemaItemNotFound { name: "Any" }', examples/dynamic_rels/main.rs:67:34
```

maldrake:


examples/dynamic_props/main.rs
```diff
@@ -26,16 +27,18 @@ fn resolve_project_points(facade: ResolverFacade<(), ()>) -> ExecutionResult {
     facade.resolve_scalar(points)
 }
 
-#[tokio::main]
```
shellforge, line 29: Curious - why the switch?

maldrake: This was the cause of #59.  When the engine, through the executor, calls the dynamic property resolver, the `ResolverFacade` creates a tokio runtime in order to interact with the Neo4J Bolt client. Because this example calls the engine executor inside a runtime context, the engine's attempt to create one causes a panic.

Once juniper releases an async version, such that the whole of the engine execute function can be wrapped in async, this should be fine.  For now, I took out the tokio execution of the main function and narrowed it down to a `block_on` call for the Neo4J client initialization.

shellforge (approved):

PR #73, opened 2020-09-19
Add option to define unauthenticated gremlin endpoint

shellforge:


src/engine/database/gremlin.rs
```diff
@@ -183,10 +185,17 @@ impl GremlinEndpoint {
         Ok(GremlinEndpoint {
             host: env_string("WG_GREMLIN_HOST")?,
             port: env_u16("WG_GREMLIN_PORT")?,
-            user: env_string("WG_GREMLIN_USER")?,
-            pass: env_string("WG_GREMLIN_PASS")?,
+            user: match env_string("WG_GREMLIN_USER") {
+                Ok(v) => Some(v),
+                Err(_) => None
```
maldrake: The function `env_string` is mostly a wrapper around `var_os` to convert the `None` value of the `Option` that `var_os` returns into an `Error` for easy error chaining with the `?` operator. It seems a little off to me to call that function and then have a `match` clause with an `Err(_)` to turn the `Error` value back into an `Option`. It's a simple enough function that I don't expect it to grow other error types over time, but even so, a wildcard branch to turn something back into what it was anyway seems sub-optimal.

In this case, I'd recommend just using `var_os("WG_GREMLIN_USR").map(|osstr| osstr.to_string_lossy().into_owned()` directly, letting it return an `Option` straight into the `user` field initializer.

src/engine/database/gremlin.rs
```diff
@@ -183,10 +185,17 @@ impl GremlinEndpoint {
...
+            },
+            pass: match env_string("WG_GREMLIN_PASS") {
+                Ok(v) => Some(v),
+                Err(_) => None
```
maldrake: Same as the comment on `user` above. Probably easier to call `var_os` and `map` directly than convert an `Option` to a `Result` and back again.

maldrake (changes requested):

maldrake (approved):

PR #75, opened 2020-09-21
WIP: added option to pass http headers to the client

drkatie:


drkatie:
@maldrake I started to add tests for this functionality, but Javi recommended against it because it would require setting up a webserver. He recommended instead just creating a WIP pull request and asking you what you think. I left a comment in the code with why I did it the way I did even though the solution isn't elegant. You might have a suggestion on a better way to do it. 

I did run the current tests (at least for neo4j) and confirmed that I didn't break anything :)

src/client/mod.rs
```diff
@@ -57,13 +60,39 @@ where
...
+            headers: {
+                let mut header_map = HeaderMap::new();
+                if let Some(headers) = headers_opt {
+                    // The solution below is not elegant, but may be the best option. A header
+                    // requires a HeaderName and HeaderValue. No matter what type you convert from
+                    // you get a Result type. For example, from_str will produce an error if the
+                    // input contains "\n" because new lines are not allowed. The code below logs
+                    // any errors that are encountered, but skips any erroneous headers in creating
+                    // the header map. I imagine we don't want the Client::Http constructor to 
+                    // return a Result type to propagate errors up.
```
maldrake: Actually `MyType::new() -> Result<MyType, Error>` is a reasonably common pattern in Rust. Fallible constructors are a thing.  I'd go that route rather than masking the error.

maldrake (commented):
In general, I'm probably in favor of testing and test coverage to a fault. Maybe it's because I write such poor quality code that I assume any line that isn't covered by a test is destined to be erroneous.  Even with that bias, though, adding a server setup and teardown in order to test what comes down to one line of the client library.  We can probably omit it in this case.

src/client/mod.rs
drkatie: I updated the constructor to return a result type and used netcat to test the ability to properly parse/send headers. No tests have been added to tests in mod.rs.

maldrake (approved):

PR #78, opened 2020-10-20
Add metadata to engine-based client and a few other utilities

shellforge:


tests/setup/extension.rs
```diff
@@ -51,9 +52,11 @@ where
     /// This hook will add metadata into the request context.
     fn pre_request_hook(
```
shellforge, line 53: Exposed two new values in this hook: the name of the operation, and access to the database.

src/client/mod.rs
```diff
@@ -921,11 +927,15 @@ where
...
+impl<G, R> Display for Client<G, R>
+where
+    G: GlobalContext,
+    R: RequestContext,
+{
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::result::Result<(), std::fmt::Error> {
         match self {
             Self::Http { endpoint, .. } => write!(f, "{}", endpoint),
-            Self::Local { engine } => write!(f, "{}", engine),
+            Self::Local { engine, .. } => write!(f, "{}", engine),
```
maldrake: Ideally `Display` ought to pretty-print the whole of a data structure, or at least as much as possible.  That would suggest pretty-printing the header map for a `Self::Http` and the metadata map for a `Self::Local`

src/engine/config.rs
```diff
@@ -419,6 +419,39 @@ impl Endpoint {
...
+    /// name: RegisterUser
+    /// class: Mutation
+    /// input:
+    ///   type: UserInput
+    /// output:
+    ///   type: User
+    /// ").unwrap();
+    /// ```
+    fn try_from(yaml: &str) -> Result<Endpoint, Error> {
+        serde_yaml::from_str(yaml).map_err(|e| Error::DeserializationFailed { source: e })
```
maldrake: Given that you added a `JsonDeserializationFailed` error elsewhere in the change, should `DeserializationFailed` be renamed `YamlDeserializationFailed` to disambiguate them?

src/engine/database/mod.rs
```diff
@@ -65,23 +65,23 @@ pub enum DatabasePool {
 
 impl DatabasePool {
     #[cfg(feature = "neo4j")]
-    pub(crate) fn neo4j(&self) -> Result<&bb8::Pool<bb8_bolt::BoltConnectionManager>, Error> {
+    pub fn neo4j(&self) -> Result<&bb8::Pool<bb8_bolt::BoltConnectionManager>, Error> {
```
maldrake, line 68: Warpgrapher should be passing a usable database client to extension points, custom endpoints, and custom resolvers that need one. If there's something broken in doing so, I think it would be best to fix that problem so that it works, if we can. I don't think we want to expose the pool creation functions and have people create their own database pools in extensions and custom resolvers.

src/engine/resolvers.rs
```diff
@@ -80,6 +81,65 @@ where
         }
     }
 
+    /// Returns the resolver input deserialized into a structure of type T that
+    /// implements the serde `Deserialize` trait.
```
maldrake, line 85: Is the idea to give people a convenience function/method that lets them work with data in a native Rust struct, rather than in serialized Warpgrapher `Value` types?

src/engine/resolvers.rs
```diff
@@ -80,6 +81,65 @@ where
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an [`Error]` variant [`InputNotFound`] if no input field was passed
+    /// to the query, [`TypeConversionFailed`] if unable to convert a warpgrapher Value
```
maldrake: I'd recommend replacing "warpgrapher Value" "[`warpgrapher::Value`]" or whatever the linked fully qualified name would be.

src/engine/resolvers.rs
```diff
@@ -80,6 +81,65 @@ where
...
+    /// to a serde_json Value, and [`JsonDeserializationFaild`] if unable to parse the
```
maldrake, line 91: I'd recommend replacing "serde_json Value" with "[`serde_json::Value`]" or whatever the correct linked fully qualified name would be.

src/engine/resolvers.rs
```diff
@@ -80,6 +81,65 @@ where
...
+    /// input data into a struct of type T.
+    ///
+    /// [`Error`]: ../../error/enum.Error.html
+    /// [`InputNotFound`]: ../../error/enum.Error.html#variant.InputNotFound
+    /// [`TypeConversionFailed`]: ../../error/enum.Error.html#variant.TypeConversionFailed
+    /// [`JsonDeserializationFailed`]: ../../error/enum.Error.html#variant.JsonDeserializationFailed
+    pub fn parse_input<T: serde::de::DeserializeOwned>(&self) -> Result<T, Error> {
```
maldrake: I'd recommend naming the function `input`, as it returns the input, but dropping parse_, as there's not really any parsing that happens in the function.

src/engine/resolvers.rs
```diff
@@ -80,6 +81,65 @@ where
...
+        let value: Value = match self.args().get("input") {
```
maldrake: You have a series of steps that each return `Some(v) => v` and/or `Ok(v) => v` and an intermediate value, `value2` that doesn't get used for anything but feeding into the next step.  Admittedly, I haven't written it out, so I may be wrong, but I would guess that with a combination of the `?` operator and some chaining functions like `and_then`, you could probably cut out some complexity.

src/engine/resolvers.rs
```diff
@@ -80,6 +81,65 @@ where
...
+        let value2 = match serde_json::Value::try_from(value) {
+            Err(_) => {
+                return Err(Error::TypeConversionFailed {
+                    src: "warpgrapher::Value".to_string(),
+                    dst: "serde_json::Value".to_string(),
+                })
+            }
+            Ok(v) => v,
+        };
+        let parsed_input: T = match serde_json::from_value(value2) {
```
maldrake: This may get rolled into a chain of calls, per the comment above, but here in particular, you're unwrapping a successful result from an Ok, only to immediately re-wrap it in an Ok and return it.  It might be simpler as:

`serde_json::from_value(value2).map_err(|e| Error::JsonDeserializationFailed {source: e})`

Or... if we have the `From` traits set up correctly for a deserialization error, it would be simpler still:

`serde_json::from_value(value2).map_err(Error::from)`

src/error.rs
```diff
@@ -59,19 +72,32 @@ pub enum Error {
     /// expected data structure
     ///
     /// [`Config`]: ../engine/config/struct.Config.html
-    DeserializationFailed { source: serde_yaml::Error },
+    DeserializationFailed {
```
maldrake: Per the other comment, given that you've introduced a `JsonDeserializationFailed`, would it make sense to rename this to avoid the appearance of overlap?  Perhaps `YamlDeserializationFailed`?

tests/setup/extension.rs
maldrake: To make sure I understand... what is the name of the operation?  Is that the name of the GraphQL top level Query or Mutation?

maldrake (changes requested):

tests/setup/extension.rs
shellforge: You can give an operation an optional name. 

Anonymous query:

```
query {
  User {
    id
  }
}
```

Named query:

```
query MyQueryForUsers {
  User {
    id
  }
}
```

src/engine/resolvers.rs
shellforge: Correct. For several reasons. One, deserializing to a struct is a good way to guarantee that you have all the fields that you are expecting in an input, and it provides a perfect way to fail fast on a request and return some type of InvalidInput error. Second, after deserializing into a struct, you can directly access fields with `mystruct.some_field` instead of `input.get("some_field").unwrap().as_string().unwrap()` except with error handling instead of unwrap.

src/engine/resolvers.rs
shellforge: Yes. I originally did it that way, and it was a very ugly looking block of code that made it hard to understand what was happening. Since the rust compiler optimizes the intermediate variable away, I figured this was a more readable approach. But I have no objection to changing it back if you think it would be preferable.

tests/setup/extension.rs
maldrake: Nice. Thank you. I have learned a new thing.

src/engine/resolvers.rs
maldrake: That makes sense.  Nice.

src/engine/database/mod.rs
shellforge: I don't think those function create new pools. I think they just return a database-specific representation of the database pool so that the user can get a client from the pool. Is that incorrect?

src/engine/resolvers.rs
maldrake: I suspect it's just stylistic preference, really, not something that would affect maintainability or, as you noted, the compiled code.  

If you keep the intermediate state and don't chain them together, it might be worth using some amount of `?`, `ok_or_else`, and `map_err` to simplify out some of the match arms that only wrap or unwrap something in an `Ok`, though.  Up to you.

src/engine/database/mod.rs
maldrake: You are totally correct. I completely misread where in the code that was and what it was doing. Let us never speak of this again...

src/client/mod.rs
shellforge: Fixed.

src/engine/config.rs
shellforge: Fixed.

src/engine/resolvers.rs
shellforge: Fixed.

src/engine/resolvers.rs
shellforge: Fixed.

src/error.rs
shellforge: Fixed.

src/engine/resolvers.rs
shellforge: Fixed.

src/engine/resolvers.rs
shellforge: Fixed.

maldrake (approved):

PR #82, opened 2020-11-09
Closes #67.

### LINKED ISSUE #67: Mismatch Between Docs and Query Syntax
https://github.com/warpforge/warpgrapher/blob/75adeaef0c43f6b77010dc652c4d262254449d55/src/engine/objects/resolvers/visitors.rs#L670-L702 looks for ADD, UPDATE, and DELETE in update queries.

The documentation book (https://warpforge.github.io/warpgrapher/api/node_update.html) says that these operations are prefixed with a dollar sign, $.

Definitely reality and documentation should match.  Based on conversations, we decided to go with adding the $, so in this case, the code should be changed to match the docs.  (Good to scan other parts of the visitors to double check anywhere else that changes to constant string operation names are needed, in case the highlighted section of code isn't the only one.)

maldrake:
Closes #67


shellforge (dismissed):

maldrake:
@shellforge, would you mind giving this a re-review now that the changes to fix test regressions are in?

shellforge (approved):

PR #91, opened 2020-12-26
Match operations

shellforge:


src/engine/database/mod.rs
```diff
@@ -18,12 +18,13 @@ use bb8_bolt::BoltConnectionManager;
...
 use gremlin_client::GremlinClient;
 use std::collections::HashMap;
+use std::convert::TryFrom;
 #[cfg(any(feature = "cosmos", feature = "gremlin", feature = "neo4j"))]
 use std::env::var_os;
 use std::fmt::Debug;
 
 #[cfg(feature = "gremlin")]
-fn env_bool(var_name: &str) -> Result<bool, Error> {
+pub fn env_bool(var_name: &str) -> Result<bool, Error> {
```
shellforge, line 27: @maldrake - is there a way to access this in `tests` without making it `pub`? I tried `pub(crate)` but it didn't work.

maldrake: I don't think so.  Integration tests are separate crates from the system under test.

src/engine/database/gremlin.rs
```diff
@@ -594,27 +594,32 @@ impl Transaction for GremlinTransaction {
...
-                query.push_str(&(".has(".to_string() + &k + ", " + &k + &param_suffix + ")"));
-            } else {
-                // For all user-defined properties, we single-quote the property name
-                query.push_str(&(".has('".to_string() + &k + "', " + &k + &param_suffix + ")"));
-            }
+        for (k, c) in props.into_iter() {
+            query.push_str(&(
+                ".has".to_string()
+                + "("
+                + if k=="id" { "" } else { "'" }  // ommit quotes if key is id because it's a "system" property
```
maldrake: Looks like a typo in the comment. Omit has one m.

src/engine/database/gremlin.rs
```diff
@@ -594,27 +594,32 @@ impl Transaction for GremlinTransaction {
...
+                + &k 
+                + if k=="id" { "" } else { "'" }  // ommit quotes if key is id because it's a "system" property
```
maldrake: Typo in omit. It should have one m.

src/engine/database/gremlin.rs
```diff
@@ -747,27 +752,32 @@ impl Transaction for GremlinTransaction {
...
-                query.push_str(&(".has(".to_string() + &k + ", " + &k + &param_suffix + ")"));
-            } else {
-                // For all other user defined properties, we sinlge quote the property name.
-                query.push_str(&(".has('".to_string() + &k + "', " + &k + &param_suffix + ")"));
-            }
+        for (k, c) in props.into_iter() {
+            query.push_str(&(
+                ".has".to_string()
+                + "("
+                + if k=="id" { "" } else { "'" }  // ommit quotes if key is id because it's a "system" property
```
maldrake, line 759: Typo in the comment. Omit has one m.

src/engine/database/gremlin.rs
```diff
@@ -747,27 +752,32 @@ impl Transaction for GremlinTransaction {
...
+                + &k 
+                + if k=="id" { "" } else { "'" }  // ommit quotes if key is id because it's a "system" property
```
maldrake, line 761: Typo in the comment. Omit has one m.

src/engine/database/gremlin.rs
```diff
@@ -1164,6 +1174,21 @@ impl TryFrom<VertexProperty> for Value {
...
+        (Operation::EQ, true) => "neq".to_string(),
+        (Operation::CONTAINS, false) => "containing".to_string(),
+        (Operation::CONTAINS, true) => "notContaining".to_string(),
+        (Operation::IN, false) => "within".to_string(),
+        (Operation::IN, true) => "without".to_string(),
+        (Operation::GT, _) => "gt".to_string(),
+        (Operation::GTE, _) => "gte".to_string(),
+        (Operation::LT, _) => "lt".to_string(),
+        (Operation::LTE, _) => "lte".to_string()
+    }
```
maldrake, line 1189: I understand why you're not using the negated flag with the mathematical operators. There's no need for not GT, if you also have LTE. But, technically, the negation is possible and well-defined. I'll try to remember to check in the schema that there's no negation flag. As long as we're not ignoring something that a user might set in a query, I have no objection here. If someone could, even if misguided on their part, actually submit a query with a negated mathematical comparison, then we should handle it correctly.

src/engine/database/mod.rs
```diff
@@ -134,6 +135,86 @@ pub trait DatabaseEndpoint {
...
+    GTE,
+    LT,
+    LTE
+}
+
+/// Struct representing a value comparison. In query operations, visitors take provided 
+/// operation/value nested map and converted them into a `Comparison` struct and pass
+/// it on the database-specific transaction for use in creating match portion of queries. 
+#[derive(Clone, Debug)]
+pub struct Comparison {
```
maldrake, line 154: FWIW, I like the way this design turned out.

src/engine/database/mod.rs
```diff
@@ -134,6 +135,86 @@ pub trait DatabaseEndpoint {
...
+
+    type Error = Error;
+
+    fn try_from(v: Value) -> Result<Comparison, Error> {
+        Ok(match v {
+            Value::String(_) => Comparison::default(v),
+            Value::Int64(_) => Comparison::default(v),
+            Value::Float64(_) => Comparison::default(v),
+            Value::Bool(_) => Comparison::default(v),
+            Value::Map(mut m) => {
```
maldrake: So, correct me if I'm wrong, but basically this lets you use the visitors to follow the input structure down to the point that you want to handle query values that may have comparisons, and then you just throw that input `Value` map at this `try_from` block and it converts everything for you?  If I understand this correctly, wow... that's more ergonomic than I expected would be possible.

src/engine/database/mod.rs
```diff
@@ -134,6 +135,86 @@ pub trait DatabaseEndpoint {
...
+                let operation_str = m.keys().next()
```
maldrake: I think what you want here is something more like

```
let (operation_str, operand) = m.into_iter().next()
```

The consuming form of the iterator, `into_iter` as opposed to `iter` (or just `keys`, as in the code now), returns owned values rather than references, avoiding the need to clone. That does consume the source, but I think that should be fine in this context, as type conversations do consume source values.

src/engine/database/mod.rs
```diff
@@ -134,6 +135,86 @@ pub trait DatabaseEndpoint {
...
+                    .ok_or(Error::ComparisonParsingFailed { message: "Comparison keys empty".to_string() })?
+                    .clone(); // TODO: is there a way to not clone here?
+                let operand = m.remove(&operation_str).unwrap();
```
maldrake: I'd recommend avoiding `unwrap` in production code. If the `into_iter` solution to obtain both key and value at the same time works, from the comment above, that solves this problem too, though.

src/engine/database/mod.rs
```diff
@@ -134,6 +135,86 @@ pub trait DatabaseEndpoint {
...
+
+    fn try_from(v: Value) -> Result<Comparison, Error> {
+        Ok(match v {
+            Value::String(_) => Comparison::default(v),
+            Value::Int64(_) => Comparison::default(v),
+            Value::Float64(_) => Comparison::default(v),
+            Value::Bool(_) => Comparison::default(v),
+            Value::Map(mut m) => {
+                let operation_str = m.keys().next()
+                    .ok_or(Error::ComparisonParsingFailed { message: "Comparison keys empty".to_string() })?
```
maldrake: Would it make sense to use the `InputItemNotFound` error here, or do you think that overloads its meaning too much?

src/engine/database/mod.rs
```diff
@@ -134,6 +135,86 @@ pub trait DatabaseEndpoint {
...
+                        _ => return Err(Error::ComparisonParsingFailed { 
+                            message: format!("Unknown operation {}", operation_str) 
+                        })
+                    },
+                    matches!(operation_str.as_ref(), "NOTEQ" | "NOTCONTAINS" | "NOTIN"),
+                    operand
+                )
+            },
+            _ => {
+                return Err(Error::ComparisonParsingFailed { message: format!("Unsupported Value type: {:#?}",  v)})
```
maldrake: Could this error be captured under the existing TypeNotExpected error? One of the examples it gives is a `Value` with an unexpected `enum` variant.  I'm looking to avoid adding additional error variants unless we really need them, but definitely push back if I'm going too far, there.

src/engine/database/neo4j.rs
```diff
@@ -465,24 +465,31 @@ impl Transaction for Neo4jTransaction<'_> {
         }
 
         if !props.is_empty() {
+            let mut value_props : HashMap<String, Value> = HashMap::new();
             props.keys().enumerate().for_each(|(i, k)| {
```
maldrake: You may be able to work off the TODO below about taking ownership by replacing this with something like

```
props.into_iter().enumerate().for_each(|(i, (k, v))| { ... })
```

src/engine/database/neo4j.rs
```diff
@@ -631,17 +638,30 @@ impl Transaction for Neo4jTransaction<'_> {
 
         let param_var = "param".to_string() + &sg.suffix();
         if !props.is_empty() {
+            let mut value_props : HashMap<String, Value> = HashMap::new();
             props.keys().enumerate().for_each(|(i, k)| {
```
maldrake: Similar to the above, I'm guessing something like

```
props.into_iter().enumerate().for_each(|(i, (k, v))| ...)
```

would allow you to take ownership and delete the line assigning to c, below.

src/engine/objects/mod.rs
```diff
@@ -547,6 +547,9 @@ where
             }
             PropertyKind::Scalar => {
                 resolver.resolve_scalar_field(info, field_name, &self.fields, executor)
+            },
+            PropertyKind::ScalarComp => {
```
maldrake: Is this a possibility? Or if it is, shouldn't it be an error?  Wouldn't ScalarComp fields only ever show up inside `Input` structs? If one ever appeared inside a `Node` struct with a resolver being called on it, wouldn't it be an error, like the `PropertyKind::Union` variant below?

src/engine/schema.rs
```diff
@@ -1865,21 +1938,144 @@ fn generate_static_version_query() -> Property {
...
+    
+    // FloatQueryInput
+    nthm.insert(
+        fmt_float_query_input_name(),
+        float_query_input()
+    );
+
     // generate graphql schema components for warpgrapher types
-    c.types().for_each(|t| {
+    //c.types().for_each(|t| {
```
maldrake: I have no complaint about switching to a for loop, if you prefer it, but there is an Iterator method try_for_each that would allow you to error chain with `?` in the closure, if that's why you made the switch.

src/engine/schema.rs
```diff
@@ -2650,7 +2851,7 @@ mod tests {
         /*
             input ProjectInput {
                 $EXISTING: ProjectQueryInput
-                $NEW: ProjectCreateMutationInput
+                NEW: ProjectCreateMutationInput
```
maldrake, line 2853: Did you mean to change this here, or did it get inadvertently pulled in from the other PR to back out the `$` prefixes from the schema?

src/error.rs
```diff
@@ -64,6 +64,11 @@ pub enum Error {
         found: i32,
     },
 
+    /// Returned when a comparison operation/value map fails to parse into a `Comparison` struct
+    ComparisonParsingFailed {
```
maldrake: Per my comments where this error is used, it may be possible to use existing error variants instead of creating a new one. However, if you think that overloads the existing errors with meanings that aren't closely enough related, I don't object to creating a new one.

src/lib.rs
```diff
@@ -12,6 +12,8 @@
 
 #![doc(html_root_url = "https://docs.rs/warpgrapher/0.6.0")]
 
+#[macro_use] extern crate maplit;
```
maldrake: If I remember right, I think macro_use is outdated syntax. You should be able to import macros from other crates using `use` statements without this kind of macro_use statement anymore.

maldrake (changes requested):
I flagged some small typo-like stuff, and I think I may have been able to help in one or two places with todo comments to avoid cloning (assuming my suggestions actually work), but I didn't have much to say.  I really like the way the design came out.

src/engine/database/mod.rs
shellforge: Yup! I was pleasantly surprised that it turned out to be this clean.

src/engine/database/mod.rs
shellforge: I knew you would know the answer :)

src/engine/database/mod.rs
shellforge: I agree on wanting to keep error variants to a minimum. I did consider `TypeNotExpected`, but it has not fields and I wanted the ability to give some context on what was happening in case we ran into issues on what is a new and significant change. I'm totally onboard with re-using `TypeNotExpected` if it has a details/message field. What do you think makes more sense?

src/engine/database/gremlin.rs
shellforge: This was something I had to iterate several times. Originally, every operation was negatable. In cypher every operation is independent and there is the ability to add a `NOT` to the entire clause. When I got to gremlin, they had a distinct operation for negations (eq/neq, within/without, etc) and not for the mathematical operations. In order to maintain consistency, I removed the negated mathematical operations from the `Operation` enum. To your point about what a user might try to do, since we are encoding operations in the schema, users won't be able to enter anything other than what is allowed in the enum.

src/engine/objects/mod.rs
shellforge: Yes - good catch.

src/engine/schema.rs
shellforge: In general when I'm mutating external state (outside the scope of the block) I prefer to use for loops - but that's a personal style preference. No objection to reverting if you prefer.

src/engine/schema.rs
shellforge: Likely accidental change as I was switching branches. All of it will be corrected in the `$` PR which is already on deck.

src/engine/schema.rs
maldrake: Nah, it's fine by me.

src/engine/database/mod.rs
maldrake: Adding a message field to the `TypeNotExpected` variant so that it better suits your purpose sounds good to me.

maldrake (approved):

PR #95, opened 2020-12-31
Event Handlers Async + Facade

shellforge:


src/engine/database/gremlin.rs
```diff
@@ -431,12 +431,12 @@ impl GremlinTransaction {
 }
 
 #[async_trait]
-impl Transaction for GremlinTransaction {
+impl<RequestCtx: RequestContext> Transaction<RequestCtx> for GremlinTransaction {
```
maldrake, line 434: Was it necessary to move the `RequestContext` from the individual function signatures into the `impl` block, or is that to reduce duplication of code?

The advantage of having the type parameter on the individual functions is that it doesn't propagate / contaminate structs that own or contain references to structs. Though I suppose that's a minor concern in this case, as request contexts show up nearly everywhere anyway. 

No strong feelings here -- this isn't a request for a change. Just want to understand the rationale.

src/engine/database/mod.rs
```diff
@@ -217,18 +231,18 @@ impl TryFrom<Value> for Comparison {
 }
 
 #[async_trait]
-pub(crate) trait Transaction: Send {
+pub trait Transaction<RequestCtx: RequestContext>: Send {
```
maldrake: Given that you create an `EventFacade` below, does this need to go from `pub(crate)` to `pub`?  Could we add a function instead that returns a `Result`-wrapped database client, instead?

Or is this a deliberate change, in line with the idea of exposing more of the visitor and `Transaction` interfaces for use by clients, to reduce the amount of direct database interaction they need? If so, are the interfaces sufficiently easy to use to be public-facing as they are?  Or do we need to make any changes, given that they weren't really built with external client use in mind?

src/engine/database/mod.rs
```diff
@@ -321,7 +335,7 @@ pub(crate) trait Transaction: Send {
 }
 
 #[derive(Clone, Debug)]
-pub(crate) struct QueryFragment {
+pub struct QueryFragment {
```
maldrake: Same question -- does this need to be made public.  Or if not needed is it desirable?

src/engine/database/mod.rs
```diff
@@ -358,7 +372,7 @@ impl QueryFragment {
 }
 
 #[derive(Clone, Debug)]
-pub(crate) struct NodeQueryVar {
+pub struct NodeQueryVar {
```
maldrake: Same question -- does this need to be made public.  Or if not needed is it desirable?

src/engine/database/mod.rs
```diff
@@ -398,7 +412,7 @@ impl NodeQueryVar {
 }
 
 #[derive(Clone, Debug)]
-pub(crate) struct RelQueryVar {
+pub struct RelQueryVar {
```
maldrake: Same question -- does this need to be made public.  Or if not needed is it desirable?

src/engine/database/mod.rs
```diff
@@ -445,7 +459,7 @@ impl RelQueryVar {
 }
 
 #[derive(Debug, Default)]
-pub(crate) struct SuffixGenerator {
+pub struct SuffixGenerator {
```
maldrake: Fixed.

src/engine/events.rs
```diff
@@ -730,3 +868,101 @@ impl<RequestCtx: RequestContext> Default for EventHandlerBag<RequestCtx> {
...
+
+    /// Returns the context of the GraphQL request which in turn contains the 
+    /// application-defined request context. 
+    pub fn context(&self) -> &'a GraphQLContext<RequestCtx> {
+        self.context
+    }
+
+    /// Returns the database transaction under which the GraphQL request is being executed.
+    /// The transaction can be used to execute database queries. 
+    pub fn transaction(&mut self) -> &mut (dyn Transaction<RequestCtx> + 'a) {
```
maldrake: Is the intent to open up the `Transaction` interface to the public here? And if so, is the DX of the interface one that works well enough, given that it was originally developed as a crate internal interface?  If not, would it make more sense to have a method that returns the client, more like the `ResolverFacade`, returning a reference to the database client within the transaction (which will thus still execute queries within the transactional context)?

src/engine/events.rs
```diff
@@ -730,3 +868,101 @@ impl<RequestCtx: RequestContext> Default for EventHandlerBag<RequestCtx> {
...
+    /// 
+    /// ```rust, no_run
+    /// # use warpgrapher::Error;
+    /// # use warpgrapher::engine::events::EventFacade;
+    /// # use warpgrapher::engine::value::Value;
+    /// # use warpgrapher::juniper::BoxFuture;
+    ///
+    /// fn before_user_update(value: Value, ef: EventFacade<()>) -> BoxFuture<Result<Value, Error>> {
+    ///     Box::pin(async move {
+    ///         let nodes_to_be_updated : Vec<Node<()>> = ef.read_nodes("User".to_string(), value, None).await?;
```
maldrake: Would this example work, as written? Doesn't it need to get the "match" portion of the update input out for the call to `read_nodes`?

src/engine/events.rs
```diff
@@ -730,3 +868,101 @@ impl<RequestCtx: RequestContext> Default for EventHandlerBag<RequestCtx> {
...
+    ///         // modify value before passing it forward ...
+    ///         Ok(value)
+    ///     })
+    /// }
+    /// ```
+    pub async fn read_nodes(
+        &mut self,
+        type_name: &str,
+        input: Option<Value>,
```
maldrake, line 978: Do we need to do anything special to document the nested Value structure we're expecting for this function, or do we just refer people to the GraphQL playground schema documentation and note that it's a serialized version of the NodeInput (at least, I think it's the GqlNodeInput... we should probably double check that).  Or, to turn that rambling thought into a helpful suggestion... I think that where we're taking in a `Value` to pass to visitors, we should document what kind of schema input node that maps to, so that people can use the playground to figure out what fields it has, and what other nested Values come underneath.

src/engine/events.rs
```diff
@@ -730,3 +868,101 @@ impl<RequestCtx: RequestContext> Default for EventHandlerBag<RequestCtx> {
...
+        info.name = "Query".to_string();
+
+        let mut sg = SuffixGenerator::new();
+        let node_var = NodeQueryVar::new(
+            Some(type_name.to_string()),
+            "node".to_string(),
+            sg.suffix(),
+        );
+
+        let query_fragment = visitors::visit_node_query_input(
```
maldrake, line 992: This design of using the facade to wrap and provide a user-friendly interface into the visitor functions seems like a great way of handling it.  (hat tip)

src/engine/objects/resolvers/mod.rs
```diff
@@ -486,30 +501,52 @@ impl<'r> Resolver<'r> {
         )
         .await?;
 
-        let results = transaction
+        // TODO: review this code block very closely
```
maldrake: As far as I can tell, this is correct.  Ugly.  But correct.  And the ugliness isn't because of you code... it's hard to clean this up given some of the language constraints. I had something that looked very clean, but I ran afoul of having to pass the transaction to the when it was already borrowed.  I'd say go with this for now and maybe we can find something more concise later.

src/engine/objects/resolvers/mod.rs
```diff
@@ -1031,33 +1079,47 @@ impl<'r> Resolver<'r> {
             transaction,
         )
         .await?;
-        let results = transaction
+
+        // TODO: review this code block very closely
```
maldrake: Looks correct to me.

src/engine/objects/resolvers/visitors.rs
```diff
@@ -400,7 +455,7 @@ where
     }
 }
 
-pub(super) fn visit_node_query_input<'a, T>(
+pub(crate) fn visit_node_query_input<'a, T: ?Sized, RequestCtx>(
```
maldrake, line 458: Why do we need to remove the Sized bound?

src/engine/objects/resolvers/visitors.rs
```diff
@@ -929,14 +1010,23 @@ where
...
-                    Ok(rels)
-                }
-            })
+            .await?;
+        if let Some(handlers) = context.event_handlers().after_rel_create(&rel_label) {
+            for f in handlers.iter() {
+                rels = f(
+                    rels,
+                    EventFacade::new(
+                        // TODO: where do we get the rel_name ?
```
maldrake: The `RelQueryVar` struct has a function to obtain it, so it'd likely be `rel_var.label()`.

src/engine/objects/resolvers/visitors.rs
```diff
@@ -955,14 +1045,27 @@ pub(super) async fn visit_rel_delete_input<T, RequestCtx>(
...
 
     let rel_label = rel_var.src().label()?.to_string() + &rel_var.label().to_title_case() + "Rel";
     let input = if let Some(handlers) = context.event_handlers().before_rel_delete(&rel_label) {
-        handlers.iter().try_fold(input, |v, f| f(v))?
+        for f in handlers.iter() {
+            input = f(
+                input,
+                EventFacade::new(
+                    // TODO: where do we get rel_name ?
+                    CrudOperation::DeleteRel(info.name().to_string(), "?".to_string()),
```
maldrake: The RelQueryVar struct has a function to obtain it, so it'd likely be rel_var.label().

src/engine/objects/resolvers/visitors.rs
```diff
@@ -986,13 +1089,26 @@ where
...
             if let Some(handlers) = context.event_handlers().after_rel_delete(&rel_label) {
-                handlers.iter().try_fold(Vec::new(), |v, f| f(v))?;
-            }
+                let mut v = Vec::new();
+                for f in handlers.iter() {
+                    v = f(
+                        v,
+                        EventFacade::new(
+                            // TODO: where do we get rel_name ?
+                            CrudOperation::DeleteRel(info.name().to_string(), "?".to_string()),
```
maldrake: The RelQueryVar struct has a function to obtain it, so it'd likely be rel_var.label().

src/engine/objects/resolvers/visitors.rs
```diff
@@ -1039,7 +1155,19 @@ where
             .await;
 
         if let Some(handlers) = context.event_handlers().after_rel_delete(&rel_label) {
-            handlers.iter().try_fold(rels, |v, f| f(v))?;
+            for f in handlers.iter() {
+                rels = f(
+                    rels,
+                    EventFacade::new(
+                        // TODO: where do we get rel_name
```
maldrake: The RelQueryVar struct has a function to obtain it, so it'd likely be rel_var.label().

src/engine/objects/resolvers/visitors.rs
```diff
@@ -1094,7 +1222,7 @@ where
     }
 }
 
-async fn visit_rel_dst_query_input<T>(
+async fn visit_rel_dst_query_input<T: ?Sized, RequestCtx>(
```
maldrake, line 1221: Same as question above... why do we need to jettison the Sized marker?

src/engine/objects/resolvers/visitors.rs
```diff
@@ -1222,7 +1351,7 @@ where
     }
 }
 
-pub(super) async fn visit_rel_query_input<T>(
+pub(super) async fn visit_rel_query_input<T: ?Sized, RequestCtx>(
```
maldrake, line 1350: Same as question above re: ?Sized.

src/engine/objects/resolvers/visitors.rs
```diff
@@ -1384,7 +1514,7 @@ where
     }
 }
 
-async fn visit_rel_src_query_input<T>(
+async fn visit_rel_src_query_input<T: ?Sized, RequestCtx>(
```
maldrake, line 1513: Same as question above re: ?Sized.

src/engine/objects/resolvers/visitors.rs
```diff
@@ -1447,7 +1578,20 @@ where
 
     let rel_label = rel_var.src().label()?.to_string() + &rel_var.label().to_title_case() + "Rel";
     let input = if let Some(handlers) = context.event_handlers().before_rel_update(&rel_label) {
-        handlers.iter().try_fold(input, |v, f| f(v))?
+        for f in handlers.iter() {
+            input = f(
+                input,
+                EventFacade::new(
+                    // TODO: where do we get rel_name ?
+                    CrudOperation::UpdateRel(info.name().to_string(), "?".to_string()),
```
maldrake: The RelQueryVar struct has a function to obtain it, so it'd likely be rel_var.label().

src/engine/schema.rs
```diff
@@ -27,7 +27,7 @@ use std::sync::Arc;
 /// [`Engine`]: ../struct.Engine.html
 #[derive(Clone, Debug, PartialEq)]
 pub struct Info {
-    name: String,
+    pub(crate) name: String,
```
maldrake, line 30: Is the pub(create) `name` function to retrieve the name inadequate?

src/engine/objects/resolvers/visitors.rs
```diff
@@ -1529,22 +1673,30 @@ where
...
-                }
-            })?;
+            .await?;
+        if let Some(handlers) = context.event_handlers().after_rel_update(&rel_label) {
+            for f in handlers.iter() {
+                rels = f(
+                    rels,
+                    EventFacade::new(
+                        // TODO: where do we get rel_name?
+                        CrudOperation::UpdateRel(info.name().to_string(), "?".to_string()),
```
maldrake: The RelQueryVar struct has a function to obtain it, so it'd likely be rel_var.label().

src/engine/database/gremlin.rs
shellforge: This came up when passing a `Transaction` to the event handler. The only way I could get it to compile was to move up the type parameter and then I had to propagate it all over the place. I definitely tried to avoid it - so if there is a way to do away with it that you can spot, I would be game for reverting it.

src/engine/database/mod.rs
shellforge: I think that the abstractions that we do/will provide on the facade will be sufficient that users do not need to use a Transaction directly. The reason it was made public is because I can't seem to get around a `can't leak crate-visible trait` compiler error. Open to ways around this so we can keep the trait non-public.

src/engine/database/mod.rs
shellforge: Fixed.

src/engine/database/mod.rs
shellforge: Fixed.

src/engine/database/mod.rs
shellforge: Fixed.

src/engine/events.rs
shellforge: Ha! I know realize this is probably the function that was causing issues with leaking the trait. This was an oversight.

src/engine/events.rs
shellforge: Fixed it.

src/engine/objects/resolvers/mod.rs
shellforge: I was hoping I had done something wrong and that there was a better way.

src/engine/objects/resolvers/visitors.rs
shellforge: This one I spent a long time fighting the compiler and this was the only way to make it work. Maybe tomorrow we can hop on for a bit and talk through any possible alternatives.

src/engine/database/mod.rs
shellforge: This was an oversight from the fact that I exposed transaction on the facade. Should be fixed now.

src/engine/objects/resolvers/visitors.rs
shellforge: Fixed.

src/engine/objects/resolvers/visitors.rs
shellforge: Fixed.

src/engine/objects/resolvers/visitors.rs
shellforge: Fixed.

src/engine/objects/resolvers/visitors.rs
shellforge: Fixed.

src/engine/objects/resolvers/visitors.rs
shellforge: Fixed.

src/engine/schema.rs
shellforge: This was a hack to be able to clone Info and "adjust" the level up to "Query" to be able to use it in `read_nodes` in the event facade. Is there a cleaner way to use Info in that situation (this is obviously a less than ideal solution).

src/engine/objects/resolvers/visitors.rs
shellforge: Fixed.

src/engine/events.rs
shellforge: Fixed.

src/engine/events.rs
shellforge: I updated the docs on this function as well as on the even handler fn type aliases.

src/engine/database/gremlin.rs
maldrake: Right now the `Transaction` reference in the `EventFacade` in a `transaction: &'a mut dyn Transaction<RequestCtx>`.  If you make that a `transaction: &'a mut T` where `T: Transaction`, does that free you up from some of the constraints, including the need to put the `RequestCtx` in the header here rather than on individual methods?

src/engine/objects/resolvers/visitors.rs
maldrake: My guess is that if you can change from a `transaction: &'a mut dyn Transaction<RequestCtx>` to a concrete type that implements Transaction, it will probably solve your sizing problem, here.  (Or we'll find out I need to learn rust better.)

src/engine/database/gremlin.rs
shellforge: I had no luck. If you find yourself with a few minutes, could you take a stab at it?

PR #96, opened 2021-01-01
Async

maldrake:
This PR makes Warpgrapher fully async.

### DIFF
--- a/examples/dynamic_rels/main.rs
+++ b/examples/dynamic_rels/main.rs
@@ -27,7 +27,7 @@ model:
        resolver: resolve_project_top_contributor
 ";
 
-fn resolve_project_top_contributor(facade: ResolverFacade<'_, ()>) -> BoxFuture<ExecutionResult> {
+fn resolve_project_top_contributor(facade: ResolverFacade<()>) -> BoxFuture<ExecutionResult> {
     Box::pin(async move {
         // create dynamic dst node
         let mut top_contributor_props = HashMap::<String, Value>::new();
--- a/examples/dynamic_rels/main.rs
+++ b/examples/dynamic_rels/main.rs
@@ -1,6 +1,5 @@
 use std::collections::HashMap;
 use std::convert::TryFrom;
-use tokio::runtime::Runtime;
 use uuid::Uuid;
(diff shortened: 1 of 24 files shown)


examples/dynamic_rels/main.rs
```diff
@@ -27,36 +27,36 @@ model:
...
-fn resolve_project_top_contributor(facade: ResolverFacade<()>) -> ExecutionResult {
-    // create dynamic dst node
-    let mut top_contributor_props = HashMap::<String, Value>::new();
-    top_contributor_props.insert(
-        "id".to_string(),
-        Value::from(Uuid::new_v4().to_hyphenated().to_string()),
-    );
-    top_contributor_props.insert("name".to_string(), Value::from("user0".to_string()));
-    let top_contributor = facade.create_node("User", top_contributor_props);
+fn resolve_project_top_contributor(facade: ResolverFacade<'_, ()>) -> BoxFuture<ExecutionResult> {
```
shellforge: @maldrake : why is a lifetime needed in the ResolverFacade here but not in the same place in the `dynamic_props` example?

src/engine/schema.rs
```diff
@@ -2307,6 +2303,11 @@ fn generate_schema(c: &Configuration) -> Result<HashMap<String, NodeType>, Error
         NodeType::new("Query".to_string(), TypeKind::Object, query_props),
     );
 
+    nthm.insert(
```
shellforge, line 2306: @maldrake : what is this type for?

examples/dynamic_rels/main.rs
maldrake: Good catch. That lifetime marking isn't needed.  I removed it in the latest push.

src/engine/schema.rs
maldrake: The new version of Juniper supports GraphQL subscriptions. The `RootNode` type has been expanded to include a root Query, root Mutation, and root Subscription. It's required for the root node, even if subscriptions are unused, so we include an empty root node with no children.

shellforge (approved):

PR #98, opened 2021-01-17
release candidate 0.7.0

shellforge:


maldrake (approved):
Oh, I should've of it back when we were working on the feature itself, but I guess we'll need a separate ticket for documentation in the book on writing event handlers.

shellforge:
@maldrake : tracking it here: https://github.com/warpforge/warpgrapher/issues/99

PR #105, opened 2021-03-07
Improve Event Facade

shellforge:


src/client/mod.rs
```diff
@@ -193,8 +192,9 @@ impl<RequestCtx: RequestContext> Client<RequestCtx> {
             Client::Local { engine, metadata } => {
                 engine
```
shellforge, line 193: Changed the signature of execute to take discrete arguments instead of requiring a juniper-specific struct (juniper::GraphQLRequest).

src/engine/config.rs
```diff
@@ -50,7 +50,7 @@ pub struct Configuration {
     ///
     /// [`Type`]: struct.Type.html
     #[serde(default)]
```
shellforge, line 52: Made the model public to allow modification in the `before_engine_build` event hook.

src/engine/config.rs
```diff
@@ -1266,6 +1266,10 @@ impl Type {
         self.props.iter()
     }
```
shellforge, line 1268: Exposed props as mutable to enable modifications during `before_engine_build` hooks.

src/client/mod.rs
```diff
@@ -193,8 +192,9 @@ impl<RequestCtx: RequestContext> Client<RequestCtx> {
...
                     .execute(
```
shellforge, line 194: Changed the signature of execute to take discrete arguments instead of requiring a juniper-specific struct (juniper::GraphQLRequest).

src/engine/config.rs
```diff
@@ -50,7 +50,7 @@ pub struct Configuration {
     ///
     /// [`Type`]: struct.Type.html
     #[serde(default)]
-    model: Vec<Type>,
+    pub model: Vec<Type>,
```
shellforge, line 53: Made the model public to allow modification in the before_engine_build event hook.

src/engine/config.rs
```diff
@@ -1266,6 +1266,10 @@ impl Type {
         self.props.iter()
     }
 
+    pub fn mut_props(&mut self) -> &mut Vec<Property> {
```
shellforge, line 1269: Exposed props as mutable to enable modifications during before_engine_build hooks.

src/engine/context.rs
```diff
@@ -69,7 +65,6 @@ pub struct GraphQLContext<RequestCtx: RequestContext> {
     resolvers: Resolvers<RequestCtx>,
     validators: Validators,
     event_handlers: EventHandlerBag<RequestCtx>,
-    extensions: Extensions<RequestCtx>,
```
shellforge, line 72: Removed extensions since they have been replaced with events.

src/engine/database/mod.rs
```diff
@@ -285,6 +286,7 @@ pub enum CrudOperation {
     UpdateRel(String, String),
     DeleteNode(String),
     DeleteRel(String, String),
+    None
```
shellforge, line 289: Added new variant since the type of crud operation is not known in the scope in which before/after request event handlers are fired.

src/engine/resolvers.rs
```diff
@@ -303,16 +311,192 @@ where
...
     ///
-    ///         let n = facade.create_node(typename, props);
+    ///         let n = facade.node(typename, props);
     ///
     ///         facade.resolve_node(&n).await
     ///     })
     /// }
     /// ```
-    pub fn create_node(&self, typename: &str, props: HashMap<String, Value>) -> Node<RequestCtx> {
+    pub fn node(&self, typename: &str, props: HashMap<String, Value>) -> Node<RequestCtx> {
```
shellforge, line 320: Renamed this method to make room with the real `create_node` method.

src/engine/objects/mod.rs
```diff
@@ -193,6 +193,32 @@ where
...
+    /// 
+    ///     // succeeds if the fields of `n` can be deserialized into `Team`
+    ///     let team: Team = n.deser().unwrap(); 
+    ///     
+    /// }
+    /// ```
+    pub fn deser<T: serde::de::DeserializeOwned>(&self) -> Result<T, Error> {
+        let m = Value::Map(self.fields().clone());
+        let v = serde_json::Value::try_from(m).unwrap();
+        let t : T = serde_json::from_value(v).unwrap();
```
maldrake: Would it make sense to replace the unwraps with errors here, since there is a `Result` return value to propagate the error through?

maldrake (commented):
Very nice.  I had one question/comment about a couple unwraps, but that was it.

In general, we've opened up a lot more surface area of APIs publicly. At some point we'll need a general tech debt cleanup of going back and both checking that those APIs really make sense and work well publicly exposed, and documenting them with public consumption in mind.

src/engine/objects/mod.rs
shellforge: Good catch.

maldrake (approved):

src/engine/objects/mod.rs
shellforge: Fixed.

PR #107, opened 2021-03-24
fixed incorrect node labels in event facades

shellforge:


shellforge:
Also ran `cargo fmt` which added a lot of noise to an otherwise straightforward PR.

maldrake (approved):

PR #109, opened 2021-04-06
added rel read function to event facade

shellforge:


src/engine/events.rs
```diff
@@ -1345,4 +1346,91 @@ where
...
+    /// # use serde_json::json;
+    /// # use warpgrapher::Error;
+    /// # use warpgrapher::engine::events::EventFacade;
+    /// # use warpgrapher::engine::objects::Rel;
+    /// # use warpgrapher::engine::value::Value;
+    /// # use warpgrapher::juniper::BoxFuture;
+    ///
+    /// fn before_user_read(value: Value, mut ef: EventFacade<()>) -> BoxFuture<Result<Value, Error>> {
+    ///     Box::pin(async move {
+    ///         let rels: Vec<Rel()> = ef.readl_rels(
```
maldrake: Typo here, I think. `readl_rels` should probably be `read_rels`.

src/engine/events.rs
```diff
@@ -1345,4 +1346,91 @@ where
...
+    pub async fn read_rels(
+        &mut self,
+        src_node_label: &str, 
+        rel_label: &str, 
+        input: impl TryInto<Value>,
+        partition_key_opt: Option<&Value>
+    ) -> Result<Vec<Rel<RequestCtx>>, Error> {
+        let input_value_opt = Some(input.try_into().map_err(|_e| Error::TypeConversionFailed {
+            src: "".to_string(),
+            dst: "".to_string(),
```
maldrake: The `dst` type here would be Value, right?  I see how there's not really a clear choice for what to put in `src`. I guess you could say `TryInto<Value>` to note that it's something that can be converted into a `Value` (maybe), but we don't know what it was.

src/engine/objects/mod.rs
```diff
@@ -210,11 +210,20 @@ where
...
     pub fn deser<T: serde::de::DeserializeOwned>(&self) -> Result<T, Error> {
+        /*
         let m = Value::Map(self.fields().clone());
         let v = serde_json::Value::try_from(m)?;
         let t: T = serde_json::from_value(v)
             .map_err(|e| Error::JsonDeserializationFailed { source: e })?;
         Ok(t)
+        */
+        let mut fields = self.fields().clone();
+        fields.insert("__label".to_string(), Value::String(self.concrete_typename.clone()));
```
maldrake, line 214: Is the idea that this corresponds to the GraphQL `__typename` attribute?  Is the name `__label` a Warpgrapher-specific convention?  Is idea is that someone writing a deserializer from a `serde_json::Value` could do something like key off of label to decide a specific `enum` variant to use, or something like that?

shellforge: The last one (it helps deserialize enums).

src/engine/events.rs
shellforge: Good catch.

src/engine/events.rs
shellforge: Fixed.

maldrake (dismissed):

src/engine/database/mod.rs
```diff
@@ -292,6 +292,7 @@ pub enum CrudOperation {
 /// Represents the different type of comparison match operations
 #[allow(clippy::upper_case_acronyms)]
 #[derive(Clone, Debug)]
+#[allow(clippy::upper_case_acronyms)]
```
maldrake, line 295: I suspect a merge artifact. This seems to duplicate two lines above.

maldrake (approved):

PR #116, opened 2021-04-25
added tracing instrumentation

shellforge:


maldrake (dismissed):
The names are inconsistent in a couple places, e.g. some of the neo4j traces end in -tx, and some don't.  Not something I'd hold up a change over, but it might be considered a breaking change if we wait to clean it up later.

shellforge:
@maldrake: naming fixed

maldrake (approved):

PR #118, opened 2021-05-16
Client ID

maldrake:
This PR:

- Adds support for client-provided IDs.
- Adds support for Neptune read replicas for read operations.
- Creates a session for Neptune mutation resolvers so that mutation queries are executed within a transaction.
-

### DIFF
--- a/src/engine/database/mod.rs
+++ b/src/engine/database/mod.rs
@@ -104,6 +104,41 @@ pub trait DatabaseEndpoint {
 pub trait DatabasePool: Clone + Sync + Send {
     type TransactionType: Transaction;
 
+    /// Returns a [`Transaction`] for the database for which this DatabasePool has connections. If
+    /// the database types offers specific read replicas, such as AWS Neptune, the connection is to
+    /// a read replica. If no separate read replicas are offered, the connection is to the same
+    /// endpoint that serves read/write requests.
+    ///
+    /// [`Transaction`]: ./trait.DatabasePool.html
+    ///
+    /// # Errors
+    ///
+    /// Returns an [`Error`] if the transaction cannot be created. The specific [`Error`] variant
+    /// depends on the database back-end.
+    ///
+    /// [`Error`]: ../../enum.Error.html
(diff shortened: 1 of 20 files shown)


src/engine/database/mod.rs
```diff
@@ -104,6 +104,41 @@ pub trait DatabaseEndpoint {
...
+    /// # #[cfg(feature = "neo4j")]
+    /// let endpoint = Neo4jEndpoint::from_env()?;
+    /// # #[cfg(feature = "neo4j")]
+    /// let pool = endpoint.pool().await?;
+    /// # #[cfg(feature = "neo4j")]
+    /// let transaction = pool.read_transaction().await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    async fn read_transaction(&self) -> Result<Self::TransactionType, Error>;
```
shellforge, line 142: Can you explain a bit more the purpose of this?

maldrake: Happy to explain.  Some of the database types we support, including Neo4J and Neptune, allow for building cluster configurations in which one node (Neptune) or multiple (Neo4J) can act as read/write nodes, and multiple other nodes in the cluster act as read-only nodes. For applications in which read operations are more prevalent than writes, this is one technique for scaling up.  For those database types, I modified the database endpoint structs to optionally take a second environment variable for a hostname for the read replicas. This assumes they're all load-balanced behind one DNS name -- we could eventually think about fancier configuration options that allow for multiple names, but I didn't go that far this time around.  Warpgrapher uses that separate name to create separate connection pools for the read-write instances vs. the read-only replicas. This method allows one to get a read-only transaction as opposed to a read-write transaction.  In hindsight, an alternative interface would've been to keep a single method for getting a transaction and pass an enum or bool specifying whether the transaction is read-only.  I can make that change if it seems cleaner.

shellforge: Makes sense.

shellforge (approved):

PR #120, opened 2021-06-01
Standard gremlin docker

maldrake:
This PR allows Warpgrapher to interoperate with the configuration of Tinkerpop provided in the `tinkerpop/gremlin-server` Docker container. The advantage is that CI/CD pipelines can pull the public container rather than having to manage a privately built configuration. The change required accommodating Tinkerpop's default of long integers for vertex and edge IDs rather than string or UUIDs. At this point, Warpgrapher just passes through whatever the client and the back-end service send to one another, without making assumptions about the type of ID being used.


.github/workflows/test.yml
```diff
@@ -21,11 +21,10 @@ jobs:
       WG_COSMOS_PORT: ${{ secrets.WG_COSMOS_PORT }}
       WG_COSMOS_USER: ${{ secrets.WG_COSMOS_USER }}
       WG_COSMOS_PASS: ${{ secrets.WG_COSMOS_PASS }}
-      WG_GREMLIN_HOST: 127.0.0.1
+      WG_GREMLIN_HOST: localhost
       WG_GREMLIN_PORT: 8182
-      WG_GREMLIN_USER: stephen
```
shellforge, line 26: @maldrake : since we are removing these, how is the user being specified? Does the `tinkerpop/gremlin-server` container use the default hard-coded username/password if no user is specified? Or are we hardcoding the user somewhere else in the code?

maldrake: The default tinkerpop/gremlin-server configuration just doesn't require user-based authentication, so no user is being specified.  We do still have test cases that use user-based auth, because it's used for Cosmos. Granted, we don't run those as part of every CI/CD run, but given that it's just passing the creds through to the gremlin-client library, I also think it's probably a low risk of regression.

shellforge (approved):

PR #132, opened 2022-02-01
fixed default value on prop required

shellforge:


src/engine/config.rs
```diff
@@ -774,7 +774,7 @@ pub struct Property {
 
     /// True if this property is required to be present on this type; false if the property is
     /// optional
-    #[serde(default = "get_false")]
+    #[serde(default = "get_true")]
```
maldrake, line 777: Needs a corresponding doc book change, but I'm good with it.

maldrake (approved):

